PR #2, opened 2015-03-08
Resolves and queues active jobs conventionally based on class name

rauhryan:
Proof of concept for conventional active job queuing

The JobResolver will first look for a job based on controller_action and then fallback to controller 

example

```
class FooController < ApplicationController
    def create
      @foo = Foo.create params
    end
end
```
would conventionally queue a `FooCreateJob`

```
#=> app/jobs/foo_create_job.rb
class FooCreateJob < ActiveJob::Base
    def perform(params)
        params[:foo] #=> the foo record
    end
end
```

:boom: :tada: :sparkles: :sparkles:

### DIFF
--- a/app/assets/javascripts/board/application.js
+++ b/app/assets/javascripts/board/application.js
@@ -2803,12 +2803,6 @@ var Repo = Ember.Object.extend(Serializable,{
   repoUrl :function () {
     return this.get("userUrl") + "/" + this.get("name");
   }.property("name", "userUrl"),
-  backlogUrl: function () {
-     return this.get("repoUrl") + "/backlog";
-  }.property("repoUrl"),
-  betaUrl: function () {
-     return this.get("repoUrl") + "/beta";
-  }.property("repoUrl"),
   fetchBoard: function(linkedBoards){
     if(this._board) {return this._board;}
     return Ember.$.getJSON("/api/" + this.get("full_name") + "/board").then(function(board){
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -29,8 +29,8 @@
       get 'issues/:number/details' => 'issues#details'
       post 'issues' => 'issues#create_issue'
       post 'issues/:number/comment' => 'issues#create_comment'
-      put 'issues/comments/:id' => 'issues#update_issue'
-      put 'issues/:number' => 'issues#update_comment'
+      put 'issues/comments/:id' => 'issues#update_comment'
+      put 'issues/:number' => 'issues#update_issue'
```
discorick: Can't believe I managed to switch those two in both the controller and the routes haha - thanks for catching that.

discorick:
Very cool dude :+1:

app/controllers/api/issues_controller.rb
```diff
@@ -67,7 +67,8 @@ def drag_card
       user, repo, number, order, column = params[:user], params[:repo], params[:number], params[:order], params[:column]
       moved = params[:moved_columns] == 'true'
       issue = huboard.board(user, repo).issue(number)
-        .move(column, order, moved)
+      @previous_column = issue['current_state']
+      @issue = issue.move(column, order, moved)
       render json: issue
```
discorick: Should probably read ```render json: @issue```

PR #23, opened 2015-03-20
rescue from the csrf failure properly

rauhryan:


app/controllers/application_controller.rb
```diff
@@ -2,12 +2,20 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+  rescue_from ActionController::InvalidAuthenticityToken, :with => :csrf_failed
+
```
discorick: Maybe instead: ``` rescue_from ActionController::InvalidAuthenticityToken, :with => :im_a_teapot ```

PR #27, opened 2015-04-01
Adds health checking to the board loading sequence

discorick:
HealthChecking lib for the back-end, comprised of DSL style 'health checks' that respond to ```perform```


lib/health_checking/health_checks/github_repo_webhooks.rb
```diff
@@ -0,0 +1,50 @@
+module HealthChecker
+  module HealthChecks
+    class GitHubRepoWebhooks
```
rauhryan: Suffix the check classes with `Check` so they are easy to search for

lib/health_checking/health_checks/github_repo_webhooks.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      ## deps
+      # {
+      #   repo: repo object,
+      #   authorization: :all, :collaborator or :admin
+      #   logged_in: bool
+      # }
+      ##
+      
+      def self.perform(deps)
```
rauhryan: if we force the checks to be instances ie;

```
def perform(deps)
    #=> stuffs
end
```
They will conform to the same constraints as ActiveJob jobs, it gives us the flexibility to offload them to a separate process 

Also makes things easy to multi-thread

lib/health_checking/health_checks/github_repo_webhooks.rb
```diff
@@ -0,0 +1,50 @@
...
+        return not_authorized unless authorized?(deps)
+        #Perform the actual Health Check
+      end
+
+      :private
+        def self.pass
```
rauhryan: See above;

lib/health_checking/checker.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    attr_reader :interface, :payload
+    def initialize(interface)
+      @payload = []
+      @interface = interface 
+    end
+
+    def check
+      checks.each do |check|
+        @payload << check.perform(@interface.deps)
```
rauhryan: RE: comment below about instance method 

`check.new.perform`

lib/health_checking/health_checks/health_check.rb
```diff
@@ -0,0 +1,51 @@
+module HealthChecks
+  module HealthCheck
+
+    def self.included(base)
+      base.extend ClassMethods
+    end
+
+    #Supports carrying over DSL attributes to the instantiated
+    #parent class
+    def initialize
```
rauhryan: initialize logic added by modules seems dangerous. 

I can think of few libraries that do it but it's always been weird to track it down.

It's just my gut, I'm willing to be wrong

lib/health_checking/health_checks/health_check.rb
```diff
@@ -0,0 +1,51 @@
+module HealthChecks
+  module HealthCheck
+
+    def self.included(base)
+      base.extend ClassMethods
```
rauhryan: RE: below

perhaps it would be better to declare class_attributes on the base class, ref: https://coderwall.com/p/xhcmbg/cattr_accessor-vs-class_attribute

```
base.class_attribute :weight
base.class_attribute :authorization
# ...
```

rauhryan: I've been trying to get better at metaprogramming to :smiley: 

I'm still learning what's best practice, I found an example in the sidekiq library that uses a similar technique

https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/worker.rb

lib/health_checking/health_checks/issue_commentable.rb
```diff
@@ -0,0 +1,48 @@
...
+    #   repo: repo object,
+    #   authorization: :all, :collaborator or :admin,
+    #   logged_in: bool
+    # }
+    ##
+    
+    def perform(deps)
+      return not_authorized unless authorized?(deps)
+      return pass if deps[:logged_in]
+      return failure
```
rauhryan: I think perform has too many responsibilities, it's doing a check and building a payload.

It seems like what you are going for is a DSL above to configure the desired payload

`name`, `weight`, `message` contains all you need to build a consistent payload. 

perform should be able to return `true` or `false`

lets assume perform **WONT** be called if there isn't sufficient authorization, at that point its comes down to a `pass` or `fail`

lib/health_checking/health_checks/health_check.rb
discorick: The sidekiq example seems decent - it's a little bit unconventional in ruby terms as attributes defined in this way act as methods, circumventing rubys internal attr getters/setters.

I'll spike it out and see how it feels.

lib/health_checking/health_checks/issue_commentable.rb
discorick: Ok yeah - I was thinking along the same lines, but you took it further (but I like)

My thoughts are another mixin (HealthCheck::Instruments?) that will facilitate the meta nature of perform/authorize behind the scenes

lib/health_checking/health_checks/health_check.rb
rauhryan: I think the sidekiq techinique sets up class_attributes to "hold" the data

If you look below a little further down the module, it sets up class methods to set the class_attributes

lib/health_checking/health_checks/issue_commentable.rb
rauhryan: That or move the payload generation to the Exam class or the Doctor

lib/health_checking/health_checks/health_check.rb
discorick: Its the other way around, the ```module ClassMethods``` methods are the dsl level methods collecting the data, ```base.class_attribute :blah``` is carrying the values forward into the parent/mixin class

rauhryan: yeah that's what I meant, sorry my fingers don't explain my brain very well :stuck_out_tongue_winking_eye:

rauhryan:
Needs UI work

rauhryan:
*a wild design appears* 

huboard

General idea being there will be a health check section for each linked repo, the fix it button will attempt to fix the issue

rauhryan:
Pretty happy with this as a good look and feel 

huboard

ember-app/app/models/health.js
```diff
@@ -0,0 +1,53 @@
+import Settings from 'app/models/settings';
+import ajax from 'ic-ajax';
+import Ember from 'ember';
+var get = Ember.get;
+
+
+var Health = Settings.extend({
+  repo: {},
+  storageKey: Ember.computed('repo.data.repo.full_name',{
```
discorick: Cool! For troubleshooting convenience, we should may add a refresh button that refetches all the checks ?

ember-app/app/models/repo.js
```diff
@@ -19,6 +20,7 @@ var Repo = Ember.Object.extend(Serializable,{
   fetchBoard: function(linkedBoards){
     if(this._board) {return this._board;}
     var self = this;
+    Health.fetch(self);
```
discorick: We are still using the old repo model?

rauhryan: Hmm might have brought that file back from a merge or something. 

Don't worry about it, it's not being used

PR #42, opened 2015-05-11
Discorick/create edit milestones

discorick:
Allows creating and editing of milestones

Special thanks to @thetimbanks for starting this, and for his large contribution to it.


discorick:
Done and Done!

ember-app/app/controllers/milestones/edit.js
```diff
@@ -0,0 +1,63 @@
...
+    var matches = _.filter(boards, board => {
+      return _.find(board.milestones, milestone => {
+        return self.get("model.originalTitle") === milestone.title;
+      });
+    });
+
+    matches.forEach(board => {
+      this.get("model")
+        .saveLinkedEdit(board, this.get("model.originalTitle"))
+        .then(function(milestone){
```
rauhryan: milestone is never used here. 

what is this block of code doing?

discorick: Milestone could get removed im surprised JShint didn't bark at me. This block is as follows:

Each match is a linked Board, We save the linked edit to the milestone of the linked board, then add the new edited milestone to that board's milestones list. This will trigger all the appropriate combinedMilestones bindings

rauhryan: jshint did bark thats how i found it.

im still unclear what you're doing when the milestone isn't being used

discorick: ```self.get("model")``` is the milestone, so technically its being used, the saveToBoard method is a milestone model method so it already has the milestone available to it.

rauhryan:
:shipit:

PR #49, opened 2015-05-18
Discorick/inclusive filtering

discorick:


discorick:
Todo:
- Query Params don't register on "straight to active" filter clicks
- Members filters still allow dim when they shouldn't

PR #54, opened 2015-05-29
hides adjacently hidden cards but leaves a little sliver for filtered ones

rauhryan:
So this was just a fun one, what do you think?

Not filtered...

rauhryan_skipping_stones_repo_huboard

Filtered...

rauhryan_skipping_stones_repo_huboard

### DIFF
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -135,6 +135,7 @@
     .active {
     }
     .filter-hidden {
+      cursor: default;
       border:0;
       padding:0;
       margin:0;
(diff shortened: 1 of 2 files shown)


discorick:
Personally I like hiding cards 100% only because busy boards tend to have slivers pushing visible cards out of view. Maybe we could "bunch up" slivers together? I think that could possibly look awesome

rauhryan:
Yep thats what it does :) 

It bunches them together, so 

this... 

rauhryan_skipping_stones_repo_huboard
vs...
rauhryan_skipping_stones_repo_huboard

discorick:
Wellllllll in that case :thumbsup:

emadrid:
noice!

PR #55, opened 2015-06-09
Fixes autoresize not being included in the app

discorick:


rauhryan:
I don't understand this PR

Where is the fix?

I see you add one file and not remove a different?

rauhryan:
Ah, was the file never included when we upgraded? lul

discorick:
Correct :D

PR #56, opened 2015-06-10
Discorick/filtering redux

discorick:
Moves filtering logic in services:

- Filters Service: Interface to control filterGroups 
- FilterGroups Service: Collects and Initializes the Filter Groups
- Member,Milestone etc. Services that implement the logic for individual groups

TODO:
- [x] Finish Search Filter
- [x] Reimplement Query Params (refactor back into subgroups)
- [x] Implement Search as its own QP


ember-app/app/mixins/issue-filters.js
```diff
@@ -0,0 +1,110 @@
...
+  //   grouping: {
+  //     labels: []
+  //   },
+  //   inclusive: {
+  //     member: [],
+  //     board: []
+  //   }
+  // }
+  //
+  filtersByMode: function(mode){
```
discorick: this is a terrible method, was only meant as a POC but I forgot to refactor. Should NOT be building complex objects n+1 times for each card render

rauhryan:
:shipit:

PR #61, opened 2015-06-22
Handles Ghee::Unauthorized exceptions

discorick:
Fixes bug where rails can lock a user out of huboard.com if their token is invalid but their csrf token is valid

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,9 +3,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
   rescue_from ActionController::InvalidAuthenticityToken, :with => :csrf_failed
-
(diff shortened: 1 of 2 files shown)


rauhryan:
@discorick excellent :) 

This saves us from throwing a 500.

What do you think about adding a redirect back to where they were when the error was thrown?

ie; 

 1. user is on ~/username/private_repo
 2. app throws Unauth
 3. user is redirect to ~/login
 4. **~/login redirects back to ~/username/private_repo** <- add this

discorick:
The only danger I see with that kind of redirect is it assumes the user will always pick the same auth level at the login screen.

rauhryan:
Let's do it, I'm pretty sure users have been asking for it for a long time

discorick:
Ok, for the JSON responses it seems like the right thing to do. For HTML redirects I am not so sure.

rauhryan:
Agreed, I think the json error is correct

PR #63, opened 2015-06-25
Weeds out duplicate repo listings

discorick:


rauhryan:
can you fix the trialing dots?

otherwise we can push this out for now

PR #71, opened 2015-07-27
Lowercases all subscribers on the socket

discorick:
When the socket fails the runloop also fails to execute all currently scheduled events - which in this case caused columns not to register to the parent controller, in turn causing crazy bugs.


discorick:
@rauhryan, want your eyes on this before we :shipit: to make sure I haven't missed something

PR #72, opened 2015-07-28
Adds event to issue activities list on close/reopen issue events

discorick:
Also removed an unused view


app/controllers/api/issues_controller.rb
```diff
@@ -19,14 +19,16 @@ def update_issue
 
     def close_issue
       user, repo, number = params[:user], params[:repo], params[:number]
-      @issue = huboard.board(user, repo).issue(number).close
-      render json: @issue
+      @issue = huboard.board(user, repo).issue(number)
+      @issue.close
+      render json: @issue.activities
```
rauhryan: Have you considered creating the equivalent event object in memory to send down to the client?

It's possible that @issue.activities could be expensive, ie; several pages long

closing/reopening issues will become slower as the amount of activities goes up

discorick: I decided against creating the object client-side because when we publish the event it would be nice for real time to update other users activities also - I suppose the incoming events could also create an in-memory activity object as well... thoughts ?

rauhryan: > I decided against creating the object client-side

I agree, I'd like the server to handle it for sure. 

> I suppose the incoming events could also create an in-memory activity object as well... thoughts ?

we could do that but it would be better to standardize on a method of doing so. I have two concerns.

 1. internally generated events could possibly duplicate events from GitHub (much like our troubles with webhook events coming in on closes/reopens done in HuBoard)
 2. code duplication on the client and server around generating activities

discorick: This change seems to also much with the issue closed publishable event

discorick:
Ok, based on conversations outside of this thread, lets delay merging this and determine whether a client-side event handler that would decorate payloads & and notify components of state changes would be a better approach

discorick:
Closing, this branch should be recut.

rauhryan:
Agreed, thank you

PR #73, opened 2015-07-29
Ensures published event channels are lowercased

discorick:
Fixes sockets sometimes subscribing to a channel but never receiving any messages


discorick:
Not ready for merge - theres still instances where sockets aren't kicking in

discorick:
@rauhryan - this should probably be packaged with the next enterprise release

rauhryan:
damn it, i thought i pushed this

discorick:
@rauhryan You pushed a client-side only version of this PR, I realized after the fact channel names should be forced lowercase server-side also

PR #74, opened 2015-07-31
Discorick/messaging

discorick:
Introduces a Messaging Mixin and an EventParser alongside the (soon to be deprecated) Socket Mixin.

Allows for events to be subscribed using the following signature:
```javascript
hbevents: {
  channel: "channel.binding",
  "type.{model.id}.action": "actionHandler",
  "{explicit.channel} type.{model.id}.action": "actionHandler"
},
_eventHandlers: {
  actionHandler: function(message){
    //Do Something
  }
}
```

TODO
- [x] Implement Publish Events for the client
- [x] Implement "_parseType" on the Event Parsing Service
- [x] Remove Socket Mixin
- [x] Test Test Test!


ember-app/app/mixins/events/issue.js
```diff
@@ -0,0 +1,31 @@
...
+  _eventHandlers: {
+    closed: function(){
+     console.log("MESSAGE RECEIVED YO");
+     //this.get("model").set("state", message.issue.state);
+    },
+    opened: function(){
+     console.log("MESSAGE RECEIVED YO");
+     //this.get("model").set("state", message.issue.state);
+    },
+  },
```
discorick: Re: ```_buildMeta```, I am not 100% sold on this pattern, essentially the messaging mixin uses it to map the local publish data, hard coding things like "type" might be too tight of a constraint

ember-app/app/controllers/issue.js
```diff
@@ -1,7 +1,15 @@
...
+import IssueEvent from "app/mixins/events/issue";
+import Messaging from "app/mixins/messaging";
 
-var IssueController = Ember.Controller.extend({
+var IssueController = Ember.Controller.extend(
+  IssueEvent, Messaging, {
   needs: ["application"],
+
+  //Subscribes to Messages in the Route after the model is set
+  subscribeDisabled: true,
```
rauhryan: remind me why this is needed?

seems very counter intuative

discorick: When the subscriber is parsing the events it looks up the identifier binding, which is based on the issue controller model. Because of the way we are instantiating this particular controller it needs to wait until the model is present before the subscriptions are set.

rauhryan: could we solve this by adding observers to the binding paths?

discorick: We wouldn't be able to normalize the event strings until all observer bindings are triggered, then send some kind of callback to trigger the event subscription - seems like things might get complex

rauhryan: could we just unsub/sub when the observer fires?

basically the same thing, ie; first pass would sub to `undefined issue.undefined.event` then when the observer fires it unsubs `undefined issue.undefined.event` and subs the bound values

probably going to be nessasary to make the issue controller work in the current state. 

my guess is there is a bug now that leaves old subscriptions when you change issues

discorick: Yah, that sounds pretty reasonable - I had thought that the controller would call ```willDestroy``` in between view changes, ill make the changes

discorick:
@rauhryan Ok I have tested the last commit fairly thoroughly, I think we are ready to :shipit:

rauhryan:
Good good!

I gave it some shit too and it seems solid

PR #78, opened 2015-08-07
Renames events to subscriptions and event handlers to subscribers

discorick:


rauhryan:
Thanks for doing that :)

Is this guy ready to :ship:?

PR #80, opened 2015-08-12
Discorick/additional susbcribers

discorick:
- [x] Linked repo issues subscribers
- [x] Milestone created subscriber and publisher
- [x] Milestone reorder susbcribers

### DIFF
--- a/app/controllers/api/milestones_controller.rb
+++ b/app/controllers/api/milestones_controller.rb
@@ -1,10 +1,10 @@
 module Api
   class MilestonesController < ApplicationController
     def create
-      milestone = huboard.board(params[:user],params[:repo])
+      @milestone = huboard.board(params[:user],params[:repo])
         .create_milestone params
 
-      render json: milestone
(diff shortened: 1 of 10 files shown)


discorick:
Not Ready for merge, reordering milestones over the wire takes far too long for the rerender on busy boards (3000ms) which bugs out when consecutive messages come in due to more than one milestone being moved. This throws the board out of sync. 

Milestone cards are taking far longer to render than they should, likely due to https://github.com/emberjs/ember.js/issues/11502, 1.13.7 seems to help but not resolve the problem.

rauhryan:
Ok, I figured at some point we would run into something like this. 

I think we can solve it by making the publisher smarter and possibly run loop aware. 

Imagine message come into through faye, we then push them onto an in memory queue that maybe starts up a run loop and publishes a message, then all other messages are blocked until the next loop. 

While you are out, I'll branch off here and spike something out. I'll ping you back when I've got something for you to take a look at.

PR #84, opened 2015-08-19
Ensures all the filter elements are model bound to computed properties

rauhryan:
[BLOCKED] - Fixes the model binding issue but breaks currently applied filters when the model bound property re-computes

Closes rauhryan/huboard#552 when merged


ember-app/app/services/filter_groups/milestone.js
```diff
@@ -1,29 +1,50 @@
...
+        }
+      }));
+      if(this._filters){
+        this._filters.forEach(function(f){
+          var newOne = filters.findBy('name', f.name);
+          if(newOne){
+            newOne.set('mode', f.mode);
+          }
+        });
+      }
```
rauhryan: Kind of YUCKY, but without doing another overhaul on the filters it's the only place to make it work

rauhryan:
ping @discorick

ember-app/app/services/filters.js
```diff
@@ -11,7 +19,8 @@ var FiltersService = Ember.Service.extend({
     if (!this.get("filterGroups.created")){ return; }
     key = key.replace("Filters", "");
     if (this.get("filterGroups." + key + ".filters")){
-      return this.get("filterGroups." + key + ".filters");
+      return this.set(key+ "Filters", attr(key));
+      return this.get(key + "Filters");
```
discorick: Line 23 will never get called due to line 22 always being returned.

ember-app/app/services/filter_groups/milestone.js
```diff
@@ -1,29 +1,50 @@
 import Ember from 'ember';
 
+function attr(modelProp, map) {
+  return Ember.computed("model." + modelProp, {
+    get: function(key){
+      var filters = this.get("model." + modelProp).map(map);
+      filters.insertAt(0, Ember.Object.create({
```
discorick: I could be missing something, but wouldn't this create a new "No Milestone" filter on every get ?

ember-app/app/services/filter_groups/milestone.js
discorick: I have never tested it, but as a callback after we create something like a milestone could we not just do something like `filters.milestoneFilters.pushObject(newMilestoneFilter)` ? It would definitely be less reactive and more dependent on knowledge of the filter internals I suppose.

ember-app/app/services/filters.js
rauhryan: Huh, good catch. 

Didn't seem to affect anything, the return value of `create` must not be significant

ember-app/app/services/filter_groups/milestone.js
rauhryan: We would need to couple the route action to the filters, which would be no bueno

ember-app/app/services/filter_groups/milestone.js
rauhryan: Yep, it's inserted every time the filters are recomputed, but it makes sense because the whole array is being mapped every time the model changes

ember-app/app/services/filter_groups/milestone.js
rauhryan: Another positive of simply observing the property is that we can make the model property observer the sort order as well.

ember-app/app/services/filter_groups/milestone.js
discorick: Right, that makes sense

ember-app/app/services/filter_groups/milestone.js
discorick: Are you referring to the order of the filters themselves in the array ?

rauhryan: I'm not sure, I didn't want to cross the work over to that part of the code base, but I was referring to making `board.filterMilestones` a "smarter" sorted array

rauhryan:
@discorick :shipit: ?

PR #89, opened 2015-08-31
Adds missing data to webhook created issues

discorick:

### DIFF
--- a/ember-app/app/mixins/subscriptions/board.js
+++ b/ember-app/app/mixins/subscriptions/board.js
@@ -12,39 +12,24 @@ var BoardSubscriptionMixin = Ember.Mixin.create({
       var issue = this.get("model.board.issues").findBy('number', message.issue.number);
       if(issue) { return issue.set("state", "open"); }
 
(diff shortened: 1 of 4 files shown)


ember-app/app/mixins/subscriptions/board.js
```diff
@@ -10,28 +10,41 @@ var BoardSubscriptionMixin = Ember.Mixin.create({
...
+    _webhookIssue: function(model){
+      var _self = this;
+      var board = this.get("model");
+      var repo = model.repo.full_name;
+      board.fetchIssue(model.number, repo).then(function(issue){
+        var newModel = Issue.create(issue);
+        if(newModel.current_state.name === "__nil__") {
+          newModel.set("current_state", _self.get("model.board.columns.firstObject"));
+        }
+
```
rauhryan: How come you don't have to find the column like you do above?

discorick: _webhookIssue cant be sure of its column until the promise resolves, so it has to be seperated

PR #93, opened 2015-09-03
Implements commit status for fullscreen PR's 

discorick:
Has dependencies on https://github.com/rauhryan/ghee/pull/33

TODO:
- [x] Style
- [x] Make Compliant with updated model


discorick:
Not yet styled.

ember-app/app/components/issue/activities/issue-status.js
```diff
@@ -0,0 +1,79 @@
...
+    this.get("status.statuses").forEach(function(status){
+      if(counts[status.state]){
+        counts[status.state] = counts[status.state] + 1;
+      } else {
+        counts[status.state] = 1;
+      }
+    });
+    return counts;
+  }.property("status.statuses.[]", "status.statuses.@each.state"),
+  aggregateState: function(){
```
discorick: This may deviate from GitHub's status policy somewhat as the `status.state` payload is by default GitHubs determined aggregateState. The difference being if any errors occur on GitHub the state is automatically determined as a failure, even if there are checks pending.

discorick:
The details link is snuggled into the description, should we push it to the right of the row ?

image

ember-app/app/components/issue/activities/issue-status.js
rauhryan: Yeah, I'm aware they consider it a failure and the API reflects that behavior. 

However, I model this behavior off of the actual behavior of the GH UI, in which case the Pending color and copy trumps all other states even when there is an error. 

I'm not sure why the have the discrepancy

rauhryan:
:shipit: ?

ember-app/app/components/issue/activities/issue-status.js
discorick: Oh interesting, I didn't realize the two didn't line up, maybe we can let the github API team know.

discorick:
:ship:

rauhryan:
Closes rauhryan/huboard#523

PR #100, opened 2015-09-24
Do not add unauthorized boards to the payload

discorick:


discorick:
Fast tracking this to get peoples boards back on line.

PR #105, opened 2015-09-29
Recomputes milestone filters on issue milestone change

discorick:
closes https://github.com/rauhryan/huboard/issues/547

### DIFF
--- a/ember-app/app/services/filter_groups/milestone.js
+++ b/ember-app/app/services/filter_groups/milestone.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
 function attr(modelProp, map) {
-  return Ember.computed("model." + modelProp, {
+  return Ember.computed("model." + modelProp, "model.issues.@each.milestoneTitle", {
     get: function(key){
       var filters = this.get("model." + modelProp).map(map);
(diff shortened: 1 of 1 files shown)


rauhryan:
Have you considered making the card recomputed its filters rather than the other way around?

Right now we got Filters -> [array of issues], maybe flip it to Issue -> [array of filters] if that makes sense 

Spider senses tell me that we are coupling filters to the "thing" being filtered

I feel like "soonish" we are going to add pulls as well as issues

I also have concerns about performance with large amounts of issues, here the filter functions would recomputed for *every* issue (I think)

discorick:
Right now we have each individual card is computing its "visible", "dim", or "hidden" state based on each filters mode. So in a sense we already have cards all watching the same array of filters and computing itself depending on filter state. But its **really expensive** as each card runs through all the active conditions and **then** determines whether its hidden, if not its runs through them all again to check if its dim, and then settles on being visible.

All this to say I agree, performance for filters has no where to go but down in the current configuration.

discorick:
Obviated by #110 . Closing.

rauhryan:
:+1:

PR #109, opened 2015-10-07
Specifies version for vagrant box to use puppet 3

efouts:
I ran into problems following the [Local Development](https://github.com/rauhryan/huboard/wiki/Local-Development) instructions. Seems to be due to being fresh and getting gems and vagrant boxes with puppet 4. The change to use environments requires some changes to the Vagrantfile, [this](https://github.com/mitchellh/vagrant/issues/3740) helped. But once, I thought I was around that I ran into other issues so this seemed like the best path to get it up and running and thought it made sense to version the box and update the wiki until someone can further investigate updating things to work with puppet 4.

In addition to the change made by this PR, I would update the wiki with these changes:

```rvm use 2.1.5```
```gem install puppet -v "<4"```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,6 +7,7 @@ VAGRANTFILE_API_VERSION = "2"
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   config.vm.box = "puppetlabs/ubuntu-14.04-64-puppet"
+  config.vm.box_version = "= 1.0.1"
 
   config.vm.synced_folder ".", "/srv/huboard", type: "nfs"
   config.vm.network "private_network", ip: "192.168.50.10"


efouts:
The different vagrant box versions can he found [here](https://atlas.hashicorp.com/puppetlabs/boxes/ubuntu-14.04-64-puppet)

discorick:
Hey @efouts , at what point did you fail with puppet >4, `vagrant up` ?

efouts:
@discorick Yes, [this](https://github.com/mitchellh/vagrant/issues/3740) was the issue during provisioning i was running into.

efouts:
@discorick @rauhryan were you able to reproduce this? I can verify that both of the machines I tried had this issue.

discorick:
@efouts I was able to reproduce the problem stemming from the incompatible box version, pinning to 1.01 looks like the right call.

 I am a little confused about the correlation with puppet and ruby version - I was able provision box 1.0.1 with Puppet 4.2.2, along with Ruby 2.2.1

efouts:
Ok, I bounced around a bunch with different errors. Let me start fresh again to verify.

efouts:
You are correct the only thing that needs to change is the vagrantfile. Ive updated this PR to be on top of master. Thanks @discorick!

discorick:
Awesome, thanks for taking the time to flush this out @efouts ! :+1:

PR #110, opened 2015-10-08
Refactors the filters implementation to decouple them from the model after the model refactor

discorick:
- Pulls out logic from the filterGroups "interface"
- Adds accessors to filters service ala `group(key)` which lazy loads a filterGroups model and creates if it has not been created (this solves a lot of weird load order problems)
- Increases the scope of task cards observes for `isFiltered`
- Removes query params `applyUrlFilters` code from the index and milestones controllers (yay!)

### DIFF
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -44,6 +44,7 @@ var Issue = Model.extend({
   },
   updateLabels : function () {
     this.set("processing", true);
+    this.set("labels", this.data.other_labels);
     return Ember.$.ajax( {
       url: `${this.get("apiUrl")}`,
       data: JSON.stringify({
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -44,7 +44,6 @@ var Issue = Model.extend({
   },
   updateLabels : function () {
     this.set("processing", true);
-    this.set("labels", this.data.other_labels);
     return Ember.$.ajax( {
       url: `${this.get("apiUrl")}`,
       data: JSON.stringify({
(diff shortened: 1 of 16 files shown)


ember-app/app/models/new/issue.js
```diff
@@ -44,6 +44,7 @@ var Issue = Model.extend({
   },
   updateLabels : function () {
     this.set("processing", true);
+    this.set("labels", this.data.other_labels);
```
rauhryan: This concerns me, labels and other_labels are very different things server-side

ember-app/app/services/filter_groups/board.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
 function attr(modelProp, map) {
-  return Ember.computed("model." + modelProp, "model." + modelProp + ".[]", {
+  return Ember.computed("model." + modelProp, {
```
rauhryan: So, the board filter won't recalculate when I add a new repo?

discorick: Line 35 sets the modelProp to `repos.[]` to begin with, it should stay the same.

ember-app/app/models/new/issue.js
discorick: This will only affect this specific issues labels array, the repo.labels are not touched - or am I missing something ?

rauhryan: issue.labels is all the labels that are on the issue

issue.other_labels is the labels minus the HuBoard labels

so I'm concerned that you are going to affect the column label later down the road by doing that

ember-app/app/services/filter_groups/board.js
rauhryan: got it

ember-app/app/models/new/issue.js
discorick: Ok, I'll take a closer look at this one

rauhryan:
cool :shipit: ?

PR #114, opened 2015-10-21
Adds authentication to all faye subscriptions

rauhryan:
Closes ZenDesk ticket 753


rauhryan:
Already deployed, fixes security hole in faye

PR #115, opened 2015-10-22
Adds sidekiq worker

rauhryan:


discorick:
Nice! Should I deploy to staging for a couple days and see how it performs ?

rauhryan:
no

PR #117, opened 2015-10-22
Uses the board rather than the assumed milestone model for a channel

discorick:


discorick:
Closing, theres a better way to fix this

PR #124, opened 2015-10-25
Fixed word wrap problem inside comment boxes

kevPo:
As you can see below, the comments on issues were not being word-wrapped, but descriptions are.  I added a class to the comments and just used the same word-break styling that was already being used for descriptions.

screen shot 2015-10-24 at 10 46 17 pm


rauhryan:
:+1: from me

discorick:
:shipit:

rauhryan:
Shipped for the world to see!

Thanks @kevPo

PR #133, opened 2015-10-29
Fixes commit references from external repos causing exceptions

discorick:


discorick:
closes #138

rauhryan:
How does this fix 605?

discorick:
It doesn't I meant to add that label to another issue - whoops!

PR #135, opened 2015-10-29
Ensures that the channel name doesn't contain dots when passed to Faye

rauhryan:
There was a bug where repos with multiples dots were throwing an exception


discorick:
:+1: , really like the move to initializers. Can't see any changes other than global replace on dotted string so I say :shipit:

PR #141, opened 2015-11-02
Ensure the commit references load the correct resource from the API

rauhryan:


ember-app/app/controllers/issue/reference.js
```diff
@@ -6,13 +6,13 @@ var IssueReferenceController = Ember.Controller.extend({
 
   fetchCommit: function(commit){
     var _self = this;
-    var repo = this.get("controllers.issue.model.repo.data.repo.full_name");
+    var repo = this.getRepoName(commit.commit_url);
     return ajax("/api/" + repo + "/commit/" + commit.commit_id)
```
discorick: Lets add `{global: false}` to this ajax call so any potential 404's don't hit the global handler.

PR #148, opened 2015-11-09
Sets up initial charge analytics with Segment

discorick:
V1 to get the data rolling in, this creates an identity per board, with a user key for each user on that board. This should allow us to create interesting data about individual boards, along with insights about usage across different boards.


app/jobs/analytics/identify_user_job.rb
```diff
@@ -0,0 +1,15 @@
+module Analytics
+  class IdentifyUserJob < AnalyticsJob
+
+    def payload(params)
+      {
+        user_id: params['owner']['id'],
+        traits: {
+          email: params['email'] || "",
+          user: params['user'],
+          account: params['account_id'] || ""
```
rauhryan: please add the account type, I would like to know whether the owner is a user or an org to correlate whether we should get rid of the user plan

config/initializers/analytics.rb
```diff
@@ -0,0 +1,5 @@
+if ENV["ANALYTICS_KEY"]
```
rauhryan: Why different env variable for Segment?

app/jobs/analytics/identify_user_job.rb
discorick: Ok - still working on what these payloads should actually be, waiting on Segment to respond.

config/initializers/analytics.rb
discorick: I didn't realize we already had a `SEGMENTIO_KEY` var.

app/jobs/analytics/identify_user_job.rb
rauhryan: specifically interested in params['user']['type']

rauhryan:
Let's talk more about extending the job resolver and queuer

I want all the analytics jobs to be in the SaaS namespace so we can exclude it from enterprise

discorick:
Ok ill hold off on further development.

rauhryan:
You're fine to keep working.

I just want to move things around before we ship

So keep doing what you're doing for now

rauhryan:
Yo @discorick check out the new hotness

I'm getting the hang of this analytics game son!

rauhryan:
@discorick :shipit:?

It's a good start to start populating `user` -> `repo owner`

discorick:
:shipit:

PR #151, opened 2015-11-18
Hides the X icon completely from the filters

rauhryan:
Three state filters are frustrating to learn. 

Experimenting with removing the X icon


discorick:
Ok, so the thinking is "if the x is not present, the user is more likely to play with the button" ?

rauhryan:
yeah, if we get a lot of push back we can figure something else out

rauhryan:
:shipit:?

discorick:
Makes sense to me :ship:

PR #153, opened 2015-11-25
Identifies unknown users as anonymous

discorick:


vendor/engines/saas/app/jobs/analytics/identify_user_job.rb
```diff
@@ -4,7 +4,7 @@ class IdentifyUserJob < AnalyticsJob
 
     def payload(params)
       {
-        user_id: params['current_user']['id'],
+        user_id: params['current_user']['id'] || "Anonymous",
```
rauhryan: I'm concerned about the repercussions of identifying anonymous users and how that will affect our intercom bill.

Please do some investigating before we commit to a strategy

discorick: In `app/views/shared/_analytics.html.erb` Line: 8 we are falling back to segments `page` method, which is pushing anonymous users down to Intercom. So, we have a couple choices:

-  Continue or Discontinue tracking anonymous page views (this will _probably_ affect whats coming into google analytics) 

- Ignore anonymous events on the backend (which actually seems prudent, as any anonymous data that could be useful is likely trackable using the client)

PR #156, opened 2015-12-03
Fixes Chrome related render bug for label overflowing label selectors

discorick:
Special thanks to @jackycute

### DIFF
--- a/app/assets/stylesheets/components/_selector.scss
+++ b/app/assets/stylesheets/components/_selector.scss
@@ -119,7 +119,7 @@
     overflow-y: auto;
     .hb-menu-item {
       padding: 6px 12px;
-      transform: translate(0);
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/components/_selector.scss
```diff
@@ -119,6 +119,7 @@
     overflow-y: auto;
     .hb-menu-item {
       padding: 6px 12px;
+      transform: translate(0);
```
rauhryan: there is another example of this in the code base here: https://github.com/huboard/huboard-web/blob/master/app/assets/stylesheets/_board.scss#L210

I'm :+1: on this fix, but I'd like to make sure we are using the mixin so that we get the net benefit of all the browser prefixing syntax

rauhryan:
:shipit:

PR #157, opened 2015-12-18
WIP: Remove use of warden and use omniauth-github in preparation for supporting GitLab

stanhu:
Relates to huboard/ghee#37.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,11 +7,11 @@ gem 'dotenv-rails', '~> 2.0.0', :require => 'dotenv/rails-now'
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
 gem 'coffee-rails', '~> 4.1.0'
+gem 'omniauth-oauth2'
+gem 'omniauth-github'
 # gem 'therubyracer', platforms: :ruby
 gem "ember-cli-rails", '0.1.13', require: nil
 gem 'ember-rails-assets'
-#gem "warden-github-rails" , github: "huboard/warden-github-rails"
-gem 'warden-github'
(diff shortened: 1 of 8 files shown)


rauhryan:
Hmmm interesting 

I will have to dig into this a little bit more when I have time. 

I ran into some problems with warden, specifically warden-github and it boiled down to SSL issues

Also, I'm not sure how we are going to support GitLab right now. 

If HuBoard should be one app or if it's completely separate "builds" of HuBoard and we have something like gitlab.huboard.com and possible github.huboard.com and so one

That's actually pretty interesting, I kind of like that idea, huboard.com could move to static site that's kind of like a train switch to the different subdomains. 

Would love to hear some thoughts on that...

rauhryan:
@stanhu I am actively looking into this right now.

Proper OAuth flow is way more complex than I thought. 

Just want to let you know that we didn't forget about you. 

Let me know if you have some free time to skype or chat about how we want to move forward with this PR

Cheers, 

 - Ryan

stanhu:
@rauhryan Somehow I missed this. Sure, maybe we can discuss this this week.

dahlbyk:
Fortunately, some of the complexity to which @rauhryan was referring was due to a quirk in GitHub `public_repo`/`repo` behavior that seems to have been resolved (see https://github.com/huboard/huboard-web/issues/172#issuecomment-196328505).

Also FYI, https://github.com/huboard/huboard-web/pull/214 adds some OAuth complexity to support deploying review apps. We'll need to implement equivalent behavior with OmniAuth.

discorick:
Closing as HuBoard for GitLab is no longer in development.

PR #161, opened 2016-01-09
[WIP] Adds configuration for server-side encrypted s3 uploads

rauhryan:


app/controllers/api/uploads_controller.rb
```diff
@@ -3,19 +3,35 @@ class UploadsController < ApiController
...
+        policy = uploader.policy do |conditions|
+          conditions << {"x-amz-server-side-encryption" => "AES256"}
+          conditions << {"x-amz-server-side-encryption-aws-kms-key-id" => ENV['AWS_KMS_KEY_ID']} if ENV['AWS_KMS_KEY_ID']
+          conditions << {'utf8' => '✓'}
+        end
+      else
+        policy = uploader.policy
+      end
+
+      uploader = {
```
discorick: It looks like Line 21 makes lines 7-9 redundant

rauhryan:
@huboard/committers I need to cut an enterprise release for this and I don't have time to wait for the official process. 


 1. merge master into enterprise
 2. merge this PR into enterprise
 3. cut new release
 4. wait for review and merge this PR into master
 5. ???? figure out how to fix the enterprise branch (help me @dahlbyk?)

I've taken WIP off, so if enough people are online to not make this happen that would be better

dahlbyk:
Just to confirm, this PR adds four environment variables:

* `AWS_ENABLED` feature toggle
* `AWS_S3_ENDPOINT` to override AWS URL (default: https://s3-us-west-2.amazonaws.com)
* `AWS_S3_ENCRYPTED` to enable encryption given `true`
* `AWS_KMS_KEY_ID` to specify that AWS should use an encryption key other than the default ([docs](http://docs.aws.amazon.com/kms/latest/developerguide/programming-keys.html))

It also adds an `ENCRYPTED_UPLOADS` feature toggle.

Really my only question is why we need both `ENCRYPTED_UPLOADS` and `AWS_S3_ENCRYPTED` - it seems we shouldn't need to require both.

ember-app/app/components/hb-markdown-composer.js
```diff
@@ -37,8 +37,15 @@ var HbMarkdownComposerComponent = Ember.Component.extend({
       fd.append('policy', response.policy);
       fd.append('signature', response.signature);
       fd.append('success_action_status', "201");
-      fd.append('file', file);
 
+      if(HUBOARD_ENV.FEATURES.ENCRYPTED_UPLOADS) {
```
dahlbyk: As more generally commented, it seems to me that this could be:

```javascript
if(HUBOARD_ENV.CONFIG && HUBOARD_ENV.CONFIG.AWS_S3_ENCRYPTED === 'true')
...
```

dahlbyk:
> Really my only question is why we need both `ENCRYPTED_UPLOADS` and `AWS_S3_ENCRYPTED` - it seems we shouldn't need to require both.

I see now that these are being populated from the same enterprise maintenance value behind the scenes. The `IMAGE_UPLOADS` feature is similarly aligned with `AWS_ENABLED`. We can :shipit: as-is, but we should revisit to settle on a single source of truth between client- and server-side.

rauhryan:
¯\_(ツ)_/¯ I wish we had more time to fix this “right"

dahlbyk:
> I wish we had more time to fix this “right"

We do before it hits `master`. So... do we need `HUBOARD_ENV.FEATURES` at all if we have `HUBOARD_ENV.CONFIG`?

PR #163, opened 2016-01-15
Improves the UX around closing issues on fullscreen issue view

discorick:
- Changes copy
- adds a close button to all modal windows

image

### DIFF
--- a/app/assets/stylesheets/_overlay.scss
+++ b/app/assets/stylesheets/_overlay.scss
@@ -77,7 +77,7 @@ body > .fullscreen-overlay {
       }
     }
   }
-  .ui-icon-x-thin {
+  .modal-close {
     position: fixed;
     color: white;
     float: right;
--- a/app/assets/stylesheets/_overlay.scss
+++ b/app/assets/stylesheets/_overlay.scss
@@ -77,7 +77,14 @@ body > .fullscreen-overlay {
       }
     }
   }
-
+  .ui-icon-x-thin {
+    position: relative;
+    color: #999;
+    float: right;
+    top: 6px;
+    font-size: 8px;
+    z-index: 1;
+  }
 }
 
 .fullscreen-card, .fullscreen-login {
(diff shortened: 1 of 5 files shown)


rauhryan:
I'd like to see a version where it's a big pretty white X in the very top right

window_and_huboard_huboard_huboard_and_iphoto

discorick:
Can do, before I do that, its worth noting this PR gives EVERY modal a closing "x" (which seems like a win). Not all of our modals dim the background like a fullscreen issue, so we would need to roll it as a "special snowflake" or provide an override in the modal view.

rauhryan:
Please list out all our modals; (screenshots if possible)

Which one's dim and which do not dim?

Perhaps the one's that don't dim are the "special snowflake" and not the other way around

discorick:
Weird... I looked through all the modals, the DO in fact dim out the background. Not sure where I got the impression otherwise. I will spike out the bold white X in the top right.

discorick:
image

@rauhryan what do you think?

rauhryan:
Yes, I approve

app/assets/stylesheets/_overlay.scss
```diff
@@ -77,7 +77,14 @@ body > .fullscreen-overlay {
       }
     }
   }
-
+  .ui-icon-x-thin {
```
rauhryan: scope this to something else, we might need to use that icon for other things within an overlay

ember-app/app/views/modal.js
```diff
@@ -36,14 +36,22 @@ var ModalView = Ember.View.extend({
...
-       this.get('controller').send('closeModal');
-     }
+     this.close();
     }.bind(this));
 
     this.$(':input:not(.close):not([type="checkbox"])').first().focus();
+
+    this.$('.modal-close').click(function(){
+      this.close();
+    }.bind(this));
```
rauhryan: Not ideal here, I think this should be an action... 

But for now, please add the unsubscribe event in the destroyElement hook

rauhryan:
:shipit:

PR #165, opened 2016-01-15
Removes a seemingly redundant toolbar

discorick:
Closes https://github.com/huboard/huboard/issues/622,

Before:
image

After:
image


rauhryan:
It's less jarring and drastic if you just remove the button

discorick:
You bet, I wanted to experiment/question the need for a toolbar in the settings context to begin with - but keeping the design between the board and the settings page seems reasonable

rauhryan:
Ok, so I'm good with removing the button

Issue #159, opened 2015-12-30
Dragging cards to issues with 0 order breaks sorting


discorick:
When dragging cards with either have 0 order, or to columns with cards with 0 order, sorting breaks and cards do not retain their expected position.

PR #170, opened 2016-01-18
Addresses issues with zero order being un-sortable (and pathogenic)

discorick:
Fix https://github.com/huboard/huboard-web/issues/159

### DIFF
--- a/lib/bridge/github/issues.rb
+++ b/lib/bridge/github/issues.rb
@@ -232,7 +232,19 @@ class Huboard
           begin
             data = MultiJson.load(match[1])
             data["order"] = self['number'] unless data["order"]
+
+            zero_adjustment = 0.1e-20
+            if data["order"] == 0
+              data["order"] = self['id'] * zero_adjustment
+              data["zero_fix"] = true
+            end
+
             data["milestone_order"] = self['number'] unless data["milestone_order"]
+            if data["milestone_order"] == 0
+              data["milestone_order"] = self['id'] * zero_adjustment
+              data["zero_fix"] = true
+            end
+
             return data
           rescue
             return { order: self['number'], milestone_order: self['number'] }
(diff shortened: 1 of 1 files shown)


dahlbyk:
Paired with @discorick on this. Had started with a client-side solution, but we can avoid that complexity by always forcing `order` and `milestone_order` to have positive, deterministic values. Two side effects from this solution:

1. `0`-order issues now sort by issue `id` instead of by repository name and issue number. Since this sort order could not have been deliberate, this seems acceptable.
2. While the new "default" sort order is intentionally very small, there is a chance that existing non-`0` issues will have a smaller `order` or `milestone_order` value, effectively injecting the `0` issues between otherwise prioritized issues. This could be mitigated by picking an even smaller value for `zero_adjustment`, but I'm content with the current value.

rauhryan:
Ok, seems legit. 

:shipit: ?

dahlbyk:
@discorick while you're testing this, can you check how we handle `order`/`milestone_order` values that are not numeric? I would expect them to get the same treatment as non-positive numeric values.

discorick:
@dahlbyk hacking the order to a string throws a 422, which is a legitimate response - it prevents the board from loading though.

discorick:
@dahlbyk review 06e2ce2 ?

lib/bridge/github/issues.rb
```diff
@@ -232,7 +232,19 @@ def embed_data(data = nil)
           begin
             data = MultiJson.load(match[1])
             data["order"] = self['number'] unless data["order"]
+
+            zero_adjustment = 0.1e-20
+            if data["order"] <= 0 || data["order"].is_a?(String)
```
dahlbyk: Seems safer to check if it's a number first, then compare to 0?

dahlbyk:
:shipit:

PR #174, opened 2016-01-21
Changes incorrect key for zero order fix with milestone orders

discorick:
Somehow this got past us, @dahlbyk quick review ?


dahlbyk:
:+1: :shipit:

PR #176, opened 2016-01-21
Removes an old CSS fix for label selector that causes breakage on newer versions of chrome

discorick:
Removes a fix for label selector scrolling the affected older versions of chrome, which now blocks newer versions of chrome


discorick:
@rauhryan or @dahlbyk  can I get a :shipit:  woop woop

dahlbyk:
For reference, this reverts https://github.com/huboard/huboard-web/pull/156 suggested by @jackycute.

:shipit:

PR #177, opened 2016-01-22
Improve HuBoard login Flow

discorick:
Closes #172 

Requirements outlined in #172

PR #181, opened 2016-01-27
UX Branch for #177 Improved Login Flow

discorick:
Moving discussion and spikes for #177 UX to this PR

### DIFF
--- a/app/assets/stylesheets/_repos.scss
+++ b/app/assets/stylesheets/_repos.scss
@@ -69,7 +69,7 @@ ul.repos li a.btn {
 
 
 .dashboard--alert {
-  h4 {
+  h5 {
     i.ui-icon-alert {
       position: relative;
       top: 1px;
--- a/app/assets/stylesheets/_repos.scss
+++ b/app/assets/stylesheets/_repos.scss
@@ -67,6 +67,16 @@ ul.repos li a.btn {
  margin-top:7px;
 }
 
+.repo-list {
+  .label {
+    &.label-success {
+      padding: 4px;
+    }
+    &.label-warning {
+      padding: 4px;
+    }
+  }
+}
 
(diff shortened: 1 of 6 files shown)


discorick:
Simple Dashboard with Private hidden on Public only Auth:

image

discorick:
@rauhryan heres an example of what we talked about this morning:
image

dahlbyk:
What if the "Looking for private repositories?" message were moved up to alert bar, e.g. instead of "Only public repositories are currently visible!" we showed:

> Looking for private reposorities? [Click here to allow private access.](#)

dahlbyk:
In fact, you could add a bit of clever CSS to have that block effectively "point" up to a disabled Private tab.

firefox_2016-01-28_11-02-20

Should be something like this...
```css
.public-only-hero:before {
  content: '';
  width: 0px; height: 0px;
  margin: -20px 25px 0px auto;
  border-bottom: 15px solid rgb(251, 238, 213);
  border-left: 15px solid transparent;
  border-right: 15px solid transparent;
}
```

discorick:
@dahlbyk So remove the left-hand button / alert altogether is what you are proposing ?

dahlbyk:
> So remove the left-hand button / alert altogether is what you are proposing ?

Yeah. Takes up the same amount of vertical space as the current alert, without the extra UI component on the left that otherwise doesn't really seem to fit. Just my $.02.

Not sure the asterisk on Private is helpful, since we don't indicate anywhere what the asterisk means.

discorick:
Asterisk is _supposed_ to indicate theres a hover over effect, it seems like its not really working. 

I actually kind of like the greyed out private tab, what do you think @rauhryan ?

discorick:
Closes in deference to #285

PR #184, opened 2016-02-01
Restore the intercom button!

discorick:


app/views/shared/_analytics.html.erb
```diff
@@ -1,9 +1,14 @@
...
   window.analytics.load("<%= ENV["SEGMENTIO_KEY"] %>");
-  window.analytics.page({ userId: '<% current_user.id %>' });
- <% end %>
+  <% if logged_in? %>
+    window.analytics.identify('<%= current_user.id %>', <%= current_user.attribs.to_json.html_safe %>)
+    window.analytics.page({ userId: '<% current_user.id %>' });
+  <% else %>
+    window.analytics.page();
+  <% end %>
```
rauhryan: We need to be very careful with the identify call

if we send null as the email field, intercom will pave over that attribute

perhaps we shouldn't send any traits at all on the client side?

app/views/shared/_analytics.html.erb
```diff
@@ -1,9 +1,14 @@
...
   window.analytics.load("<%= ENV["SEGMENTIO_KEY"] %>");
-  window.analytics.page({ userId: '<% current_user.id %>' });
- <% end %>
+  <% if logged_in? %>
+    window.analytics.identify('<%= current_user.id %>', <%= current_user.attribs.to_json.html_safe %>)
+    window.analytics.page({ userId: '<% current_user.id %>' });
```
dahlbyk: This should be redundant with `identify` restored. Maybe just reset to exactly what we had before?

discorick:
Ok, it might just be in our best interest to start using https://docs.intercom.io/install-on-your-web-product/use-javascript-to-install-intercom-on-any-web-site on the client side.

I am gonna look into it further.

rauhryan:
> Ok, it might just be in our best interest to start using https://docs.intercom.io/install-on-your-web-product/use-javascript-to-install-intercom-on-any-web-site on the client side.
> 
> I am gonna look into it further.

I don't want this to hold up simply getting the button to reappear, can we start a new thread to discuss getting Acquire installed

discorick:
Alrighty! After some local testing it appears passing in a blank object in client-side `identify` calls does not trigger any kind of property overwrite with the intercom integration :+1:

dahlbyk:
:shipit:

PR #185, opened 2016-02-01
Moves over the modal-close x to account for scroll bars

discorick:
image


dahlbyk:
What does it look like if there isn't a scroll bar?

discorick:
No scrollbar

image

dahlbyk:
It's weird that `right` isn't respecting the scroll bar, but I don't think it really matters for now.

:shipit:

PR #191, opened 2016-02-02
Guard against missing emails

### LINKED ISSUE #190: Login Error If Emails Not Found
The login jobs introduced by #182 are failing periodically with:

> no implicit conversion of String into Integer
>
> /app/vendor/engines/saas/app/jobs/login_private_job.rb:5 []
> /app/vendor/engines/saas/app/jobs/login_private_job.rb:5 block in perform
> /app/vendor/engines/saas/app/jobs/login_private_job.rb:5 each
> /app/vendor/engines/saas/app/jobs/login_private_job.rb:5 detect
> /app/vendor/engines/saas/app/jobs/login_private_job.rb:5 perform

The failing users show a Not Found payload instead of the expected array -

```json
          emails: {
            "message": "Not Found",
            "documentation_url": "https://developer.github.com/v3"
          }
```

dahlbyk:
Fixes #190

### DIFF
--- a/vendor/engines/saas/app/jobs/login_private_job.rb
+++ b/vendor/engines/saas/app/jobs/login_private_job.rb
@@ -1,7 +1,7 @@
 class LoginPrivateJob < ActiveJob::Base
 
   def perform(params)
-    if params['emails']
+    if params['emails'].respond_to?('detect')
       email = params['emails'].detect{|e| e['primary'] == true}
       params['user']['email'] = email['email']
       params['user']['email_verified'] = email['verified']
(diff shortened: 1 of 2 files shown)


rauhryan:
Did we get caught by the strange 404 edge case?

:shipit:

dahlbyk:
> Did we get caught by the strange 404 edge case?

Indeed. I _assume_ the 404 is happening if a user somehow avoids re-authenticating with `user:email`. The 404 payload from GitHub is then passed through by Ghee and we're treating it as an `Enumerable`. I'm a little confused why we didn't see this in testing when we removed `user:email` from the scopes required.

Having slept on this, I'm now not 100% sure the `respond_to?` check is sufficient... The "no implicit conversion of String into Integer" error doesn't seem to suggest that `detect` is failing, but rather that we're incorrectly using the element within the block.

PR #193, opened 2016-02-02
Actually guard against missing emails

### LINKED ISSUE #190: Login Error If Emails Not Found

dahlbyk:
Actually fixes #190. The `respond_to?` check added by #191 doesn't actually exclude a `Hash`.

### DIFF
--- a/vendor/engines/saas/app/jobs/login_private_job.rb
+++ b/vendor/engines/saas/app/jobs/login_private_job.rb
@@ -1,7 +1,7 @@
 class LoginPrivateJob < ActiveJob::Base
 
   def perform(params)
-    if params['emails'].respond_to?('detect')
+    if params['emails'].is_a?(Array)
       email = params['emails'].detect{|e| e['primary'] == true}
       params['user']['email'] = email['email']
       params['user']['email_verified'] = email['verified']
--- a/vendor/engines/saas/app/jobs/login_public_job.rb
+++ b/vendor/engines/saas/app/jobs/login_public_job.rb
@@ -1,7 +1,7 @@
 class LoginPublicJob < ActiveJob::Base
 
   def perform(params)
-    if params['emails'].respond_to?('detect')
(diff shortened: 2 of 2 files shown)


vendor/engines/saas/app/jobs/login_private_job.rb
```diff
@@ -1,7 +1,7 @@
 class LoginPrivateJob < ActiveJob::Base
 
   def perform(params)
-    if params['emails'].respond_to?('detect')
+    if params['emails'].is_a?(Array)
       email = params['emails'].detect{|e| e['primary'] == true}
       params['user']['email'] = email['email']
```
discorick: This will throw if `email` is an empty array, I suppose we can do something like `params['emails'].is_a?(Array) && params['emails'].size`

dahlbyk: What about `params['emails'].detect{|e| e['primary'] == true} || {}`?

rauhryan:
haha, ok 

sorry for jumping the gun

:shipit:

vendor/engines/saas/app/jobs/login_private_job.rb
discorick: That would also work !

Issue #179, opened 2016-01-26
[SAAS] Goal tracking added to Google Analytics


rauhryan:
Build up analytic goal tracking for HuBoard

- [x] Enterprise Contact/Trial submit
- [x] Enterprise Login to Dashboard
- `shared/analytics` in every view
   - [x] `/login` #194
   - [ ] Errors?
- Login actions
   - [x] `/login/(public|private)` ✓ #232
   - [x] `/login/(public|private)/authorized` ✓ #232
- Basic account actions
   - [x] `/settings/{owner}/trial` when email prompt is shown ✓
   - [x] `/settings/{owner}/trialing` when email prompt is submitted ✓ #207 
   - [x] `/settings/{owner}/expired` when an expired trial board is accessed ✓ #208 
   - [x] `/settings/{owner}/canceled` when a cancelled account board is accessed ✓ #208
- Account reactivation (requirement of https://github.com/huboard/huboard/issues/635)
   - `/settings/{owner}/reactivate` when a long-expired/cancelled board is accessed
   - `/settings/{owner}/reactivating` when a reactivation trial is started
- Board actions
   - [x] `/{owner}/{repo}/board/create` when we offer to create HB labels ✓
   - [x] `/{owner}/{repo}/board/creating` when we are asked to create HB labels ✓ #233
- `/settings/profile` actions
   - [x] `/settings/{owner}/upgrade` when Upgrade is clicked (popping Stripe)
   - [x] `/settings/{owner}/upgrading` when Stripe is successfully submitted (`/settings/charge/dahlbyk`) ✓ #230
   - [x] `/settings/{owner}/canceling` when an account is cancelled ✓ #230
   - What else?

dahlbyk:
Ideas for specific goals:

1. Login Success (New Auth or Returning User)
2. Paywall => Trial Prompt => Trial Signup
3. "Trial Expired" Paywall => Account Upgrade
4. Trial Account (not Expired) => Account Upgrade
5. Enterprise Trial Form => Start Trial

rauhryan:
I must be confusing "Goals" with "segmenting"

I think I get those list items with intercom segmentation

dahlbyk:
Think of a Goal as a funnel - in traditional ecommerce it might be Landing Page :point_right: Product Page :point_right: Cart :point_right: Checkout :point_right: Confirmation.

[...]

PR #194, opened 2016-02-02
Adds analytics to the login view

discorick:
requirement 3 of #179


dahlbyk:
So far, so good. Do you want to add to the error views with this PR, too?

discorick:
Yeah, that would make sense - I am going to put this PR on ice while I finish out the `/settings/trials` analytics

rauhryan:
Simple enough... :shipit:

PR #196, opened 2016-02-03
Guard against missing primary email

dahlbyk:
Re: https://github.com/huboard/huboard-web/pull/193#discussion_r51589794


discorick:
:shipit:

PR #198, opened 2016-02-08
Adds word-break: break-all; to card titles in order to prevent columns from growing beyond the limit

rauhryan:
related to huboard/huboard#631

### DIFF
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -303,6 +303,7 @@
   }
   .title {
     word-wrap: break-word;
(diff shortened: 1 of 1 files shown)


discorick:
:shipit:

dahlbyk:
This was a step backward - all words wrap wherever now, not just words that are too long to fit.

seanlinsley:
I'd have to agree, it's much harder to read ticket titles now.

rauhryan:
huboard_huboard_huboard

seanlinsley:
:+1:

PR #199, opened 2016-02-08
Adds word-break to issue fullscreen as well

rauhryan:
huboard/huboard#631

### DIFF
--- a/app/assets/stylesheets/_issue.scss
+++ b/app/assets/stylesheets/_issue.scss
@@ -57,6 +57,7 @@
 }
 
 .flex-issue-title {
+  word-break: break-all;
   @include display(flex);
   .issue-closed-badge p {
     margin: 4px 4px;
--- a/app/assets/stylesheets/_issue.scss
+++ b/app/assets/stylesheets/_issue.scss
@@ -57,7 +57,7 @@
 }
 
 .flex-issue-title {
-  word-break: break-all;
+  @include word-break(break-all);
   @include display(flex);
   .issue-closed-badge p {
     margin: 4px 4px;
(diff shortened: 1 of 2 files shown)


rauhryan:
rauhryan_skipping_stones_repo_huboard

rauhryan:
rauhryan_skipping_stones_repo_huboard

dahlbyk:
Can we include the vendor prefixes? From https://css-tricks.com/almanac/properties/w/word-break/:

```css
 -ms-word-break: break-all;
     word-break: break-all;

     // Non standard for webkit
     word-break: break-word;

-webkit-hyphens: auto;
   -moz-hyphens: auto;
        hyphens: auto;
```

dahlbyk:
Also, http://stackoverflow.com/a/1795878/54249 seems to suggest `word-wrap: break-word` might be more correct here. https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/ has a more complete discussion, with this CSS:

```css
.dont-break-out {

  /* These are technically the same, but use both */
  overflow-wrap: break-word;
  word-wrap: break-word;

  -ms-word-break: break-all;
  /* This is the dangerous one in WebKit, as it breaks things wherever */
  word-break: break-all;
  /* Instead use this non-standard one: */
  word-break: break-word;

  /* Adds a hyphen where the word breaks, if supported (No Blink) */
  -ms-hyphens: auto;
  -moz-hyphens: auto;
  -webkit-hyphens: auto;
  hyphens: auto;

}
```

rauhryan:
```
word-wrap: break-word
```

didn't work, if we see it pop up again we can reopen

dahlbyk:
:shipit:

PR #200, opened 2016-02-09
Remove unnecessary bower_components

dahlbyk:
Added by @rauhryan in 6d8a8e2fa9dbb115fd709abd52cad39504a0af71, but the files don't seem to be referenced anywhere?


discorick:
:shipit:

Issue #187, opened 2016-02-01
Fullscreen breadcrumbs for linked issues is not working from the parent repository


rauhryan:
![uploads/833fd8a5-4646-4770-be22-8409973c3ed2/huboard_huboard_HuBoard.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F833fd8a5-4646-4770-be22-8409973c3ed2%2Fhuboard_huboard_HuBoard.png)

Steps to reproduce:

  1. Open huboard/huboard
  2. Find issue huboard/huboard-web#186
  3. Notice the fullscreen breadcrumbs is not correct

Possible causes: 

  * Columns from linked repositories doesn't rely on the hardened array index but instead on the user entered label name index. 
  * Columns are based on a string comparison?

![uploads/b6a3a408-dd0a-4cb9-96fb-3f42294abe09/Labels_·_huboard_huboard-web_and_Labels_·_huboard_huboard_png.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fb6a3a408-dd0a-4cb9-96fb-3f42294abe09%2FLabels_%C2%B7_huboard_huboard-web_and_Labels_%C2%B7_huboard_huboard_png.png)

Expected results:

See picture above, give that both repositories have the same column names, we should treat them equally based on `real index`

discorick:
Looks like a WIP problem with linked repositories, after syncing WIP's between repos heres how it looks:
 
![uploads/5b0703d1-6db3-42e8-9546-dc56a8ba1db2/huboard_huboard_HuBoard.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F5b0703d1-6db3-42e8-9546-dc56a8ba1db2%2Fhuboard_huboard_HuBoard.png)

rauhryan:
Is it possible to make it work regardless of the string name

PR #204, opened 2016-02-16
Fixes missing status on column breadcrumbs

discorick:
Addresses #187 

Column crumb statuses were computing based on their names, however this breaks down on WIP columns when a linked repo does not share the same WIP. Fixes by moving a columns sortedIssues into the column itself, enabling us to check if the issue itself exists in the column or not.


ember-app/app/components/hb-column-crumb.js
```diff
@@ -6,10 +6,10 @@ var HbColumnCrumbComponent = Ember.Component.extend({
   classNameBindings: ["stateClass", "isSelected:active:inactive", "indexClass"],
   isSelected: function(){
     if(this.get("issue.data.state") === "closed"){
-      return this.get("visibleColumns.lastObject.data.name") === this.get("column.data.name");
+      return this.get("visibleColumns.lastObject.data.index") === this.get("column.data.index");
     }
-    return this.get("issue.data.current_state.name") === this.get("column.data.name");
-  }.property("issue.data.current_state.name", "column.data.name", "issue.data.state"),
+    return this.get("issue.columnIndex") === this.get("column.data.index");
+  }.property("issue.columnIndex", "column.data.index"),
```
rauhryan: Is this based on 'real' index or 'user input' index

`'real' index` == position in the array of columns
`user input` index == `X - Column name` from label

discorick: That would be 'user input' index

discorick: But I can see where you are going with that, I am going to reimplement using real array index.

ember-app/app/components/hb-column-crumb.js
```diff
@@ -5,11 +5,11 @@ var HbColumnCrumbComponent = Ember.Component.extend({
...
-    if(this.get("issue.data.state") === "closed"){
-      return this.get("visibleColumns.lastObject.data.name") === this.get("column.data.name");
-    }
-    return this.get("issue.data.current_state.name") === this.get("column.data.name");
-  }.property("issue.data.current_state.name", "column.data.name", "issue.data.state"),
+    var current_issue = this.get("issue");
+    return this.get("column.sortedIssues").find((issue)=> {
+      return issue.get("id") === current_issue.get("id");
+    });
+  }.property("column.sortedIssues.@each"),
```
rauhryan: How does this affect the render time on the milestones view?

ember-app/app/models/new/column.js
```diff
@@ -6,7 +6,37 @@ var Column = Ember.Object.extend(Ember.PromiseProxyMixin, {
...
+    return this.get("columns.firstObject.data.name") === this.get("data.name");
+  }.property("columns.firstObject"),
+  sortedIssues: function(){
+    var issues = this.get("board.issues")
+      .filter(this.filterStrategy.bind(this))
+      .sort(this.sortStrategy);
+    return issues;
+  }.property("board.issues.@each.{columnIndex,order,state}"),
+  filterStrategy: function(issue){
+    var issue_index = issue.get("columnIndex");
```
dahlbyk: To make sure I'm reading this right, this is the only actual logic change within the relocated column functions (was `var issue_index = issue.data.current_state.index`). Everything else is just adjustment to add/remove `model`?

ember-app/app/components/hb-column-crumb.js
```diff
@@ -5,11 +5,11 @@ var HbColumnCrumbComponent = Ember.Component.extend({
...
   classNameBindings: ["stateClass", "isSelected:active:inactive", "indexClass"],
   isSelected: function(){
-    if(this.get("issue.data.state") === "closed"){
-      return this.get("visibleColumns.lastObject.data.name") === this.get("column.data.name");
-    }
-    return this.get("issue.data.current_state.name") === this.get("column.data.name");
-  }.property("issue.data.current_state.name", "column.data.name", "issue.data.state"),
+    var current_issue = this.get("issue");
+    return this.get("column.sortedIssues").find((issue)=> {
+      return issue.get("id") === current_issue.get("id");
```
dahlbyk: Is there any measurable efficiency to be gained from capturing `current_issue.get("id")`?

```javascript
var current_id = this.get("issue").get("id");
return this.get("column.sortedIssues").find((issue)=> {
 return issue.get("id") === current_id;
});
```

And another Ember newbie question: would `this.get("issue.id")` work?

ember-app/app/models/new/column.js
discorick: Yes, it was just a way to keep that particular piece of data consistent, if the logic on the computed property `columnIndex` ever changes it's one less potential breakage

ember-app/app/components/hb-column-crumb.js
discorick: I doubt it, if anything it would be incredibly minuscule. The Ember syntax for that line would be more accurately:

`var current_id = this.get("issue.id");` 

I personally find writing methods with comparisons in the callback easier to read if the left and the right are as similar is possible.

To answer your final question this block _could_ also be written like so:

```javascript
return this.get("column.sortedIssues").find((issue)=> {
 return issue.get("id") === this.get("issue.id");
}.bind(this));
```

Which has a nice terseness to it but the `bind` method tends to trip people (and some versions of phantomjs) up.

ember-app/app/components/hb-column-crumb.js
discorick: I double checked the rendering speed with the Ember inspector, they are virtually identical. 

I am interested when I make the sister change with the milestone column model if we see any change. But that is a seperate PR.

ember-app/app/models/new/column.js
dahlbyk: :cool: Just trying to separate the refactoring from other changes.

ember-app/app/components/hb-column-crumb.js
dahlbyk: Yeah, not a fan of the `bind` version. But thanks for clarifying that `get("foo").get("bar")` is equivalent to `get("foo.bar")`.

I assume a simple `get()` would be hyper-optimized, but thought I'd ask.

dahlbyk:
This change looks good to me. :shipit:

rauhryan:
:shipit:

rauhryan:
Any reason this hasn't been deployed yet?

discorick:
Nope, just deployed!

PR #207, opened 2016-02-19
Adds analytics for trial created

discorick:
- Implements the page method for `Analytics::Core`
- Pages on trial created

References #179


drusellers:
So, HuBoard submits data to GA via a delayed job?

discorick:
@drusellers In this case that is correct - there's certain actions we actually want to track as page visits to get a better understanding of how HuBoard is being used (and how we can make it better), any of these "meta-visits" I Page using a delayed job.

vendor/engines/saas/app/jobs/analytics/page_job.rb
```diff
@@ -0,0 +1,13 @@
+module Analytics
+  class PageJob < AnalyticsJob
+    action "page"
+
+    def payload(params)
+      {
+        user_id: params['current_user']['id'] || "Anonymous",
+        name: params['url'],
+        properties: { url: params['url'] }
+      }
```
rauhryan: `user_id: params['current_user']['id'] || "Anonymous",`

I think this will throw if current_user is nil, I think you need to check both

vendor/engines/saas/app/jobs/saas/trials_create_job.rb
```diff
@@ -0,0 +1,10 @@
+module Saas
+  class TrialsCreateJob < ActiveJob::Base
+
+    def perform(params)
+      user = params['current_user']['login']
+      params['url'] = "/settings/#{user}/trialed"
+      Analytics::PageJob.perform_later(params)
+    end
```
rauhryan: Same as above? Could blow up if current_user is nil?

vendor/engines/saas/app/jobs/analytics/page_job.rb
discorick: Yah, that needs to be handled better thanks for the catch

vendor/engines/saas/app/jobs/saas/trials_create_job.rb
discorick: In this case if `current_user` is nil I would hope for an exception to help me trace where trial sign-ups are going wrong. Seem reasonable ?

rauhryan: As long as it doesn't get caught in a retry loop and spam us

dahlbyk: Is it correct to be using `login` here if a trial is being initiated for an organization?

dahlbyk:
FWIW the other `AnalyticsJob`s all use `user_id: params['current_user']['id'] || "Anonymous"`. I haven't noticed any errors from it, but I would support adding your safer version as `AnalyticsJob#user_id` to be used on all of them. Or just move setting `user_id` down into `AnalyticsJob#perform`.

dahlbyk:
> So, HuBoard submits data to GA via a delayed job?

When sending to Segment.io from Ruby, yes. See complete breakdown in https://github.com/huboard/huboard-web/issues/179#issuecomment-176260182.

We also send data (in particular normal `identify` and `page` events) client-side.
firefox_2016-02-19_16-01-38

vendor/engines/saas/app/jobs/saas/trials_create_job.rb
discorick: It's the only way for GA goal tracking to follow the order of events no?

dahlbyk: I didn't phrase that well - if you try to access an org repo rather than a user repo, does it track `/settings/{org}/trialed` or `/settings/{user}/trialed`? `current_user.login` suggests the latter.

discorick: It's /settings/{login}/trialed for both, I was under the assumption we want to track the subscriber/owner

rauhryan:
@discorick @dahlbyk 

Just a quick thought, but

Doesn't `/settings/trialed/{login}` play better with content drilldown in GA?

I could easily drill through to `(trialed|trialing|expired)` and get a quick count and overview with minimal effort

discorick:
Good question, we have just been matching on our current URL scheme. The create trial notification lives at `/settings/{user}/trial` and so I have been matching everything else to that. I am not sure about whats easier for content drilldown though.

vendor/engines/saas/app/jobs/saas/trials_create_job.rb
dahlbyk: > It's /settings/{login}/trialed for both, I was under the assumption we want to track the subscriber/owner

We can identify the subscriber/owner from the actual payment information. In GA I _think_ it's more useful to know what user/org gained or lost access.

This would also align with `/settings/{org}/trial`, correct?

dahlbyk:
> Doesn't `/settings/trialed/{login}` play better with content drilldown in GA?
> 
> I could easily drill through to `(trialed|trialing|expired)` and get a quick count and overview with minimal effort

The goal is to provide your quick count and overview through Goals, plus conversion rate and funnel paths.

--------------------

On an unrelated note, your use of `trailing` here is interesting. I had proposed `trialed`, but I think `triailing` might be a better choice to differentiate from the account "states" of expired/canceled. The flow would be:

* `trial` :point_right: `triailing` :point:right: `expired`
* `upgrade` :point_right: `upgrading` :point_right: :`canceled`
* `reactivate` :point_right: `reactivating` :point_right: `expired`

Our goals would be `trialing`, `upgrading` and `reactivating`.

vendor/engines/saas/app/jobs/saas/trials_create_job.rb
```diff
@@ -0,0 +1,10 @@
+module Saas
+  class TrialsCreateJob < ActiveJob::Base
+
+    def perform(params)
+      user = params['current_user']['login']
+      params['url'] = "/settings/#{user}/trialing"
```
dahlbyk: Please apply the `owner` change here as well.

dahlbyk:
> Track new trials based on repo owner

This is looking good. Go ahead and kick the tires on the review app to make sure you're seeing what you expect for this and #208 in the Segment dev bucket. If you're satisfied, go ahead and :shipit:.

PR #208, opened 2016-02-19
Adds Analytics for Subscription Canceled and Trial Expired

discorick:
- Splits out unauthenticed action on `Saas::ErrorController` to `unauthenticated_canceled` & `unauthenticated_expired` (along with discrete views)
- Adds conventional jobs for both actions
- Pages on Subscription Canceled and Trial Expired 

References #179

### DIFF
--- a/lib/hu_board/account_helpers.rb
+++ b/lib/hu_board/account_helpers.rb
@@ -10,6 +10,10 @@ module HuBoard
       customer && customer[:trial] == "active"
     end
 
+    def trial_expired?(customer)
+      customer && customer[:trial] == "expired"
+    end
+
     def non_profit?(customer)
       customer[:stripe][:plan] && 
         customer[:stripe][:plan][:plan_id] == "non_profit"
@@ -21,6 +25,15 @@ module HuBoard
       if cus[:subscriptions][:total_count] > 0
         sub = cus[:subscriptions][:data][0]
         return sub[:status] == "active" || sub[:status] == "trialing"
+      end
+      false
+    end
+
+    def subscription_canceled?(customer)
+      cus = customer[:stripe][:customer]
+      if cus[:subscriptions][:total_count] > 0
+        sub = cus[:subscriptions][:data][0]
+        return sub[:status] == "canceled"
       end
       false
     end
--- a/lib/hu_board/account_helpers.rb
+++ b/lib/hu_board/account_helpers.rb
@@ -35,7 +35,7 @@ module HuBoard
         sub = cus[:subscriptions][:data][0]
         return sub[:status] == "canceled"
       end
-      false
+      true
     end
 
     def account_exists?(customer_doc)
--- a/vendor/engines/saas/app/jobs/saas/errors_unauthenticated_canceled_job.rb
+++ b/vendor/engines/saas/app/jobs/saas/errors_unauthenticated_canceled_job.rb
@@ -2,8 +2,8 @@ module Saas
   class ErrorsUnauthenticatedCanceledJob < ActiveJob::Base
 
     def perform(params)
-      user = params['current_user']['login']
-      params['url'] = "/settings/#{user}/canceled"
+      owner = params['action_controller.params']['user']
+      params['url'] = "/settings/#{owner}/canceled"
       Analytics::PageJob.perform_later(params)
     end
   end
--- a/vendor/engines/saas/app/jobs/saas/errors_unauthenticated_canceled_job.rb
+++ b/vendor/engines/saas/app/jobs/saas/errors_unauthenticated_canceled_job.rb
@@ -1,0 +1,10 @@
+module Saas
+  class UnauthenticatedCanceledJob < ActiveJob::Base
+
+    def perform(params)
+      user = params['current_user']['login']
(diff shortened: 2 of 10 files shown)


discorick:
Heres how the views are looking:

<b>Owner/Admin:</b>

Canceled
image

fevish:
<b>Collaborator:</b>
collab-canceled

rauhryan:
This PR reminds me of the warden failure app and that I haven't thought about that with the new OmniAuth stuff.

dahlbyk:
> Collaborator: Expired

This matches Canceled - should it?

lib/hu_board/account_helpers.rb
```diff
@@ -25,6 +29,15 @@ def subscription_active?(customer)
       false
     end
 
+    def subscription_canceled?(customer)
+      cus = customer[:stripe][:customer]
+      if cus[:subscriptions][:total_count] > 0
+        sub = cus[:subscriptions][:data][0]
+        return sub[:status] == "canceled"
```
dahlbyk: Confirming "cancelled" vs "canceled" here, as I've seen related bugs in the past.

discorick: Stripe uses "canceled" verses "cancelled", so I have been maintaining that convention in the app.

vendor/engines/saas/app/jobs/saas/errors_unauthenticated_expired_job.rb
```diff
@@ -0,0 +1,10 @@
+module Saas
+  class ErrorsUnauthenticatedExpiredJob < ActiveJob::Base
+
+    def perform(params)
+      user = params['current_user']['login']
+      params['url'] = "/settings/#{user}/expired"
+      Analytics::PageJob.perform_later(params)
+    end
+  end
```
dahlbyk: Same question: https://github.com/huboard/huboard-web/pull/207#discussion_r53524045

vendor/engines/saas/app/jobs/saas/errors_unauthenticated_canceled_job.rb
```diff
@@ -0,0 +1,10 @@
+module Saas
+  class ErrorsUnauthenticatedCanceledJob < ActiveJob::Base
+
+    def perform(params)
+      user = params['current_user']['login']
+      params['url'] = "/settings/#{user}/canceled"
+      Analytics::PageJob.perform_later(params)
+    end
+  end
```
dahlbyk: Same question: https://github.com/huboard/huboard-web/pull/207#discussion_r53524045

vendor/engines/saas/lib/saas/before_action.rb
```diff
@@ -25,6 +25,14 @@ def check_account
           end
 
           failure :unauthorized do
+            if params[:subscription_canceled]
+              throw :warden, action: 'unauthenticated_canceled'
+            end
+
+            if params[:trial_expired] && !params[:subscription_canceled]
```
dahlbyk: `&& !params[:subscription_canceled]` here should be unnecessary since we already threw for that case.

dahlbyk:
Two other seemingly conflicting comments:

1. Can you set up a third job to track `ErrorsController#unauthenticated_saas` as `/settings/{user}/unauthenticated` so we can see if any cases are slipping through the cracks?

2. From our discussion yesterday I was kind of expecting this to be modeled closer to how `/settings/{user}/trial` is handled (a redirect to a real route), rather than actions on `ErrorsController` and fake pages tracked via job. The benefit of actually taking the user to the expired/canceled page is that we'd be able to more easily start A/B testing those pages to see what leads to improvements in upgrade conversion rate. Thoughts?

lib/hu_board/account_helpers.rb
dahlbyk: :cool: :+1:

discorick:
> Can you set up a third job to track `ErrorsController#unauthenticated_saas`

That shouldn't be a problem

>   I was kind of expecting this to be modeled closer to how `/settings/{user}/trial` is handled (a redirect to a real route)

That was my original intent, I can't seem to get Rails to make a URL transition off the back of warden throwing, I spent a little time digging into it further and couldn't seem to come up with any real solution other than treating all of these routes as redirect_to's instead .

dahlbyk:
> > Can you set up a third job to track `ErrorsController#unauthenticated_saas`
> 
> That shouldn't be a problem

:+1: 

> > I was kind of expecting this to be modeled closer to how `/settings/{user}/trial` is handled (a redirect to a real route)
> 
> That was my original intent, I can't seem to get Rails to make a URL transition off the back of warden throwing, I spent a little time digging into it further and couldn't seem to come up with any real solution other than treating all of these routes as redirect_to's instead .

That's fine, as long as it's tracking consistently for now. We can unify the approach the next time we make changes here.

----

I think the only open question here is the user/login decision: https://github.com/huboard/huboard-web/pull/207#discussion_r53524045.

discorick:
Ok! I think this just about does it, @dahlbyk shall we :ship: ?

dahlbyk:
:clap:

PR #210, opened 2016-02-23
Implementing App Enabled Emails

discorick:
References https://github.com/huboard/corp/issues/48

- Adds a Mailer class in `vendor/engines/saas/lib`
- Adds a Mailer Adapter for Sendgrid
- Uses rspec for unit testing (the TestCase framework looses its mind on the Rails Engine ENV)

TODO:
- CouchDB docs for individual accounts indicating if it's their first login or not
- Jobs for scheduling the emails with sideqik


discorick:
@rauhryan @dahlbyk , Wanted to have some discussion around how we want to structure our couchdb docs - going to need to start saving docs for individual users to coordinate sending out helpful emails.

PR #211, opened 2016-02-23
Implements App Enabled Emails

discorick:
Merges into `discorick/canceled_and_expired_analytics`
References https://github.com/huboard/corp/issues/48

The Scope of this this PR changed quite a bit during its life-time. From where we stand today this accomplishes the following:

- [x] Sets up `Action::Mailer` in the Saas Engine with some _very_ basic layouts
- [x] Configures mailer previews and injects an interceptor to allow attachment previews (for things like images)
- [x] Adds a new user document in couchdb for every OAuth'd user for future opportunities to communicate with them  
- [x] Adds mailcatcher and a puppet provisioner for it


discorick:
@rauhryan @dahlbyk , Wanted to have some discussion around how we want to structure our couchdb docs - going to need to start saving docs for individual users to coordinate sending out helpful emails.

vendor/engines/saas/app/jobs/base_login_job.rb
```diff
@@ -0,0 +1,36 @@
...
+    payload = map_payload(params)
+    Analytics::IdentifyUserJob.perform_later(payload)
+
+    #email = map_email(params)
+    #Queue job to send the email
+  end
+
+  def self.action(action)
+    @action = action
+  end
```
rauhryan: I remember having trouble subclassing class attributes 

Why aren't you using the class_attr stuff here?

discorick: Not 100% following you. This is the same pattern we have been using with the Analytics Jobs, and the Issue Jobs for Child Class instance vars.

rauhryan: We use this pattern in the healthchecks

https://github.com/huboard/huboard-web/blob/master/lib/health_checking/health_checks/health_check.rb#L5-L40

I remember there was a reason for it

config/environments/development.rb
```diff
@@ -1,4 +1,12 @@
 Rails.application.configure do
+  #Mailer Previews
+  config.action_mailer.preview_path = Saas::Engine.root.join('test/mailers')
```
dahlbyk: Help me understand what this is doing.

puppet/manifests/base.pp
```diff
@@ -8,9 +8,13 @@
   'npm': ensure => "present";
   'couchdb': ensure => "present";
   'libffi-dev': ensure => "present";
+  'libsqlite3-dev': ensure => "present";
   } ->
   class { 'ruby_install': 
     ruby_version => "2.2.1",
+  } ->
+  exec { 'mail-catcher':
+    command => 'gem install mailcatcher'
```
dahlbyk: It strikes me as odd that this is the only `gem install`. Why is mailcatcher special?

vendor/engines/saas/Gemfile.lock
```diff
@@ -42,25 +43,26 @@ GEM
...
-    multi_json (1.10.1)
+    minitest (5.7.0)
+    multi_json (1.11.2)
     nokogiri (1.6.6.2)
       mini_portile (~> 0.6.0)
-    rack (1.6.0)
+    rack (1.6.4)
     rack-test (0.6.3)
       rack (>= 1.0)
     rails (4.2.0)
```
dahlbyk: Alarm bells are ringing, thus #237.

config/environments/development.rb
discorick: This line points the `Action::Mailer` previewer to looks for mailer previews in the `Saas::Engine`. Otherwise when we attempt to use previews Rails is not able to find them.

puppet/manifests/base.pp
discorick: Mail-catcher has dependencies that make it incompatible with a Gemfile install. It runs as a daemon, and requires a different build of event-machine than Rails does. I did attempt to bundle it with all the other gems and the Daemon broke.

dahlbyk: That's odd, but :ok:

dahlbyk:
Pulled `master` in due to `BaseLoginJob` overlap in #232 - I told Git to use what's in `master`, but you might want to double-check: https://github.com/huboard/huboard-web/commit/f2795e2795525476759036034e8af8c2279869f9#diff-409415fd755afb7ec1e211a3c1b9b519

vendor/engines/saas/Gemfile.lock
discorick: Gotcha, thanks for opening that!

discorick:
@dahlbyk `f2795e2#diff-409415fd755afb7ec1e211a3c1b9b519` did go a little bit sideways. I am not sure the how go about fixing it.

dahlbyk:
> f2795e2#diff-409415fd755afb7ec1e211a3c1b9b519 did go a little bit sideways. I am not sure the how go about fixing it.

How does https://github.com/huboard/huboard-web/commit/9746411a531f6b61ab846fa32928f29e0ed55835#diff-409415fd755afb7ec1e211a3c1b9b519 look?

dahlbyk:
> dahlbyk closed this 3 minutes ago

Oops :imp:

firefox_2016-03-15_22-57-07

discorick:
Should I just reopen this against master at this point ?

dahlbyk:
> Should I just reopen this against master at this point ?

Yup.

PR #212, opened 2016-02-25
Fixes link tab being broken

discorick:
When the prior "isSelected" bug was fix by recomputing based on an issue, it was neglected that the link page also uses the same component. This is now broken when isSelected tries to compute and cannot find and issue.


rauhryan:
:shipit:

PR #201, opened 2016-02-09
Clean Vagrant+Windows Install

dahlbyk:
Here are some miscellaneous changes from getting Vagrant up and running on a clean Windows machine.

1. `.gitattributes` for EOL consistency
2. Remove `pwd` (for debugging) from `postinstall` (Windows doesn't like the `&&`)
3. Further adjust how `postinstall` calls `bower` to be more Windows-friendly - @discorick can you confirm `cd ember-app && npm install` works as expected if you remove their cache dirs?
4. Add <del>`provision.sh` for extra provisiony stuff, starting with adding</del> Puppet `exec` to add `cd /srv/huboard` to `.bashrc`
5. <del>Force downgrade bundler to 1.9.6.</del>
   - <del>It feels like there should be a better way to do this, but I can't tell exactly where the `bundler` gem is being installed.</del>
   - <del>@rauhryan, are you able to fill in background on why newer versions won't work?</del>

PR #214, opened 2016-02-26
Implement Heroku Review Apps

### LINKED ISSUE #183: Continuous Delivery per Pull Request on Heroku
https://devcenter.heroku.com/articles/github-integration-review-apps

dahlbyk:
Closes #183. As [discussed](https://github.com/huboard/huboard-web/issues/183#issuecomment-181741222), due to OAuth host constraints this isn't as trivial as it would be for a simple app.

- [x] Add `app.json` review app config, mostly generated by Heroku.

    I did manually add `HEROKU_PARENT_APP_NAME`, which I wish they had included in the generated file. It took me a _long_ time to figure out that `app.json` only matters when a review app is first created - this second commit didn't take effect until I deleted the existing review app. Much time wasted here.

- [x] Override [`Warden::GitHub::Config.normalized_uri()`](https://github.com/atmos/warden-github/blob/3e9aa5fa55d52a0cf8d0cd69ecfb07ce1d5fc03a/lib/warden/github/config.rb#L115-L122) to do two things:

    1. Update `host` with the parent app name instead of the current app name.
    2. Append an `APP_NAME` query string with the current app name, so we can redirect back to it.

- [x] <del>Update `LoginController`</del> Add middleware to check for `APP_NAME` so we can redirect to review apps from the parent app as necessary.

### DIFF
--- a/app.json
+++ b/app.json
@@ -51,6 +51,9 @@
     "HEROKU_APP_NAME": {
       "required": true
     },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": true
+    },
     "HEROKU_RELEASE_VERSION": {
       "required": true
     },
--- a/app.json
+++ b/app.json
@@ -1,0 +1,127 @@
+{
+  "name": "huboard-web",
+  "scripts": {
+  },
+  "env": {
+    "AWS_ACCESS_KEY_ID": {
+      "required": true
+    },
+    "AWS_S3_BUCKET": {
+      "required": true
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+      "required": true
+    },
+    "BUILDPACK_URL": {
+      "required": true
+    },
+    "CLOUDFRONT_URL": {
+      "required": true
+    },
+    "COUCH_DATABASE": {
+      "required": true
+    },
+    "COUCH_URL": {
+      "required": true
+    },
+    "ELASTICSEARCH_URL": {
+      "required": true
+    },
+    "GITHUB_API_ENDPOINT": {
+      "required": true
+    },
+    "GITHUB_AUTH_STRATEGY": {
+      "required": true
+    },
+    "GITHUB_CLIENT_ID": {
+      "required": true
+    },
+    "GITHUB_SECRET": {
+      "required": true
+    },
+    "GITHUB_WEB_ENDPOINT": {
+      "required": true
+    },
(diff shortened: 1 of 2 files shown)


dahlbyk:
A good chunk of experimentation happened over in #201. Here's what it looks like so far:
firefox_2016-02-26_00-30-19

dahlbyk:
:boom: https://huboard-rails-pr-214.herokuapp.com/ will now round-trip auth through https://huboard-rails.herokuapp.com/.
firefox_2016-02-27_13-10-35

Code is all in one file for hackability, but would welcome guidance on how best to structure it relative to the existing app. And general feedback, of course.

Steven-Pingel:
:clap:

dahlbyk:
For reference, when you auth from a PR build you should see the Warden Override in the logs (e.g. `heroku -t -a huboard-rails-pr-214`):

```
2016-03-03T17:48:24.910837+00:00 app[web.1]: Warden Override: huboard-rails-pr-214 => huboard-rails
2016-03-03T17:48:24.910902+00:00 app[web.1]: Warden Override: https://huboard-rails.herokuapp.com/login/public?APP_NAME=huboard-rails-pr-214
```

And then on the parent side (e.g. `heroku -t -a huboard-rails`) the app redirect is handled:

```
2016-03-03T17:48:24.864070+00:00 app[web.1]: AppRedirect: huboard-rails => huboard-rails-pr-214
2016-03-03T17:48:24.865453+00:00 app[web.1]: AppRedirect to http://huboard-rails-pr-214.herokuapp.com/login/public?code=d438whatever0342&state=7a88ce7e378cf5afwhatever56854fcdce9b5
```

Looking again at the logs, I'm going to suppress the `AppRedirect` logging if `app_name` is not set. We're seeing a lot of `AppRedirect: huboard-rails =>` noise.

dahlbyk:
@discorick raises a good point - the `app.json` will need to be maintained as we add new environment variables. Also note that existing review apps need to be destroyed and recreated to if their `app.json` changes environment variable settings - the copy happens when the review app is created.

discorick:
@dahlbyk talked in length about where the code in the initializer should live (i.e in the bridges middleware stack), ultimately we decided to leave the Warden code where it is. This seems like the right course of action currently as we look into moving to OmniAuth #157

PR #216, opened 2016-03-03
fixes a bug with coupon code redemption

### LINKED ISSUE #215: Coupon codes can not be redeemed

rauhryan:
Closes #215 when merged

### DIFF
--- a/vendor/engines/saas/app/controllers/saas/coupons_controller.rb
+++ b/vendor/engines/saas/app/controllers/saas/coupons_controller.rb
@@ -11,7 +11,7 @@ module Saas
     def redeem
       query = Queries::CouchCustomer.get_cust(params[:id], couch)
       doc = QueryHandler.exec(&query)
-      return render json: {success: false, message: "Couldn't find couch record: #{doc.id}"} unless doc
+      return render json: {success: false, message: "Couldn't find couch record: #{params[:id]}"} unless doc
 
       begin
         customer = Stripe::Customer.retrieve(params[:id])
--- a/vendor/engines/saas/app/controllers/saas/coupons_controller.rb
+++ b/vendor/engines/saas/app/controllers/saas/coupons_controller.rb
@@ -11,8 +11,8 @@ module Saas
     def redeem
       query = Queries::CouchCustomer.get_cust(params[:id], couch)
       doc = QueryHandler.exec(&query)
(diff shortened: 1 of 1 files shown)


vendor/engines/saas/app/controllers/saas/coupons_controller.rb
```diff
@@ -11,8 +11,8 @@ def valid
     def redeem
       query = Queries::CouchCustomer.get_cust(params[:id], couch)
       doc = QueryHandler.exec(&query)
-      return render json: {success: false, message: "Couldn't find couch record: #{plan_doc.id}"} unless doc
       plan_doc = doc
+      return render json: {success: false, message: "Couldn't find couch record: #{plan_doc.id}"} unless doc
```
dahlbyk: So if `doc` is `nil`, then `plan_doc` will also be `nil` and `plan_doc.id` will throw...?

dahlbyk: To be more constructive, it seems like the error message needs to reference `params[:id]`.

rauhryan: Yep, so I just moved the line down. 

could be refactored to just use doc I suppose

dahlbyk: But the line is only called `unless doc` - if it's hit, `doc` won't have an `id`.

discorick: `plan_doc` is `false` if doc fails `unless` -

+1 to "To be more constructive, it seems like the error message needs to reference params[:id]" from @dahlbyk

vendor/engines/saas/app/controllers/saas/coupons_controller.rb
```diff
@@ -11,16 +11,15 @@ def valid
     def redeem
       query = Queries::CouchCustomer.get_cust(params[:id], couch)
       doc = QueryHandler.exec(&query)
-      return render json: {success: false, message: "Couldn't find couch record: #{plan_doc.id}"} unless doc
-      plan_doc = doc
+      return render json: {success: false, message: "Couldn't find couch record: #{doc.id}"} unless doc
```
dahlbyk: Using `doc` alone now is :cool: but I still think we need to use `params[:id]` for the error message.

PR #217, opened 2016-03-03
Bumps rails from 4.2.0 -> 4.2.4

dahlbyk:
Pulling this out of #211 so we can :ship: it separately.


Gemfile.lock
```diff
@@ -282,8 +282,8 @@ GEM
       net-ssh (>= 2.6.5)
     net-ssh (2.9.2)
     netrc (0.10.3)
-    nokogiri (1.6.6.2)
-      mini_portile (~> 0.6.0)
+    nokogiri (1.6.7.2)
+      mini_portile2 (~> 2.0.0.rc2)
```
dahlbyk: Any idea why this is `2.0.0.rc2` when version is `2.0.0` above?

discorick:
Ok @dahlbyk I am pretty confident wit this one now - I ran into a problem with arguments not serializing/deserializing between ActiveJobs and it took me way to long to figure out what was wrong. 

Essentially plain `Hash`'s are now being transported as such, before this upgrade they were being automatically coalesced into `HashWithIndifferentAccess` 

See c02ed1a

seanlinsley:
Y'all should upgrade to 4.2.5.2 soon. It includes a number of security fixes.

dahlbyk:
> See c02ed1a

:+1: 

> Y'all should upgrade to 4.2.5.2 soon. It includes a number of security fixes.

@discorick thoughts on doing this now while we're already bumping versions?

discorick:
@seanlinsley @dahlbyk I would like to get this guy merged in first to unblock some PR's in flight. But point taken, +1 for security fixes, we should look into 4.2.5.2 as soon as we can

dahlbyk:
> But point taken, +1 for security fixes, we should look into 4.2.5.2 as soon as we can

Issue created. Thanks for the tip @seanlinsley!

PR #158, opened 2015-12-18
Rewrite the Ember Accounts app using Ember 2.2

iheanyi:
Moving the accounts app away from Ember-Tools and rewriting it using Ember-CLI and Ember 2.2.

Closes #167 when merged

PR #230, opened 2016-03-10
Adds analytics for upgraded & canceled subscriptions

discorick:
`/settings/{owner}/upgrade` when Upgrade is clicked (popping Stripe modal) is blocked by #158/#189/#209

### DIFF
--- a/vendor/engines/saas/app/jobs/saas/plans_destroy_job.rb
+++ b/vendor/engines/saas/app/jobs/saas/plans_destroy_job.rb
@@ -2,8 +2,8 @@ module Saas
   class PlansDestroyJob < ActiveJob::Base
 
     def perform(params)
-      if params[:event] == 'plan_canceled'
-        params['url'] = "/settings/#{params[:repo_owner]}/canceled"
+      if params[:event] == 'plan_canceling'
+        params['url'] = "/settings/#{params[:repo_owner]}/canceling"
         Analytics::PageJob.perform_later(params)
       end
     end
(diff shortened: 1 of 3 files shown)


dahlbyk:
Random observation: `/settings/profile` doesn't have a payload for Identify, just the ID. Probably fine since we associate info with that ID elsewhere, but wanted to point it out since it surprised me.
image

discorick:
Is that the JS or Ruby call @dahlbyk ?

vendor/engines/saas/app/jobs/saas/plans_destroy_job.rb
```diff
@@ -0,0 +1,11 @@
+module Saas
+  class PlansDestroyJob < ActiveJob::Base
+
+    def perform(params)
+      if params[:event] == 'plan_canceled'
+        params['url'] = "/settings/#{params[:repo_owner]}/canceled"
```
dahlbyk: This should be `canceling`; `canceled` is for hits on a board that has already been canceled.

dahlbyk:
> Is that the JS or Ruby call @dahlbyk ?

Both, actually.
image

vendor/engines/saas/app/jobs/saas/plans_destroy_job.rb
discorick: Your right, I mixed up on the verbiage

dahlbyk:
Looking good.
image

Issue #172, opened 2016-01-20
Improve Login Flow


discorick:
Good enough:
  - [x] Change the copy of the "Sign in" button to an icon button with the octocat and "Sign in with GitHub" - @discorick 
  - [x] /login Simplify the two step process with just [Sign in with GitHub] - @discorick 
  - [x] /login Add (Green icon) Your email to the permissions asked for - @discorick 
  - [x] "Sign in with github button" Change it's default behavior to "public" scope + email - @discorick 
  - [x] Force a re-auth for all _new_ "login" clicks
  - [ ] Move public vs private to separate OAuth applications - @rauhryan
  - [ ] Store credentials in couchdb (must be encrypted)
  - [ ] Change copy on pricing page to say we store tokens with XYZ encryption and never look at your data without your expressed permission
  - [x] Ensure that email that is collected is making it into intercom [#1 criteria for acceptance] 
  - [x] Sense public auth only on the dashboard and provide an upgrade option - @discorick 

Better:
  - [ ] Skip /login all together and default to "Email scope" 
  - [ ] [DASHBOARD]  If we only have the "email" scope, the landing page should be a "Welcome" here's how you get started similar to what [/login] does today. ![uploads/749cceb6-3b13-4550-9416-94f083f33aa3/huboard_huboard-web_172_pdf__1_page_.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F749cceb6-3b13-4550-9416-94f083f33aa3%2Fhuboard_huboard-web_172_pdf__1_page_.png)
  - [x] [DASHBOARD] If we know they only have Public access, have an option to upgrade to private ("Not seeing your private repositories? [Upgrade access] [?]"

rauhryan:
@discorick if you need further inspiration or ideas, refer to https://libraries.io for general UX ideas

drusellers:
> Store credentials in couchdb (must be encrypted)

Ohhh!!! :+1:

drusellers:
I'd love to review the encryption approarch

rauhryan:
keep an eye out for the PR ;)

dahlbyk:
> [Sign in with github button] Change it's default behavior to "public" scope + email

Note that explicitly requesting public scope will actively remove permissions for private scope. You had mentioned having multiple applications for OAuth, does that come into play here? Or is there a way with Warden to just make sure we have at least `read:org` (and `user:email`?) to pass through to the repo list?

[...]

PR #232, opened 2016-03-11
Adds analytics for Login/Scope

discorick:
This should help us understand better if there are any issues/objections in our login funnel when users attempt to authorize via GitHub

### DIFF
--- a/vendor/engines/saas/lib/saas/engine.rb
+++ b/vendor/engines/saas/lib/saas/engine.rb
@@ -23,7 +23,7 @@ module Saas
         :private
 
         def page_job
-          if request.referer !~ /github\.com/
+          if request.referer !~ /github\.com/ && !logged_in?
             Analytics::PageJob.perform_later({
               'url' => "/login/#{params['action']}"
             })
--- a/vendor/engines/saas/lib/saas/engine.rb
+++ b/vendor/engines/saas/lib/saas/engine.rb
@@ -17,6 +17,19 @@ module Saas
         })
       end
 
+      LoginController.class_eval do
+        before_action :page_job, only: [:public, :private]
+
+        :private
+
+        def page_job
+          if request.referer !~ /github\.com/
+            Analytics::PageJob.perform_later({
+              'url' => "/login/#{params['action']}"
+            })
+          end
+        end
+      end
     end
   end
 end
--- a/vendor/engines/saas/lib/saas/engine.rb
+++ b/vendor/engines/saas/lib/saas/engine.rb
@@ -23,7 +23,8 @@ module Saas
         :private
 
         def page_job
-          if request.referer !~ /github\.com/ && !logged_in?
+          auth_request = request.params['code'] && request.params['state']
+          if request.referer !~ /github\.com/ && !logged_in? && !auth_request
             Analytics::PageJob.perform_later({
               'url' => "/login/#{params['action']}"
             })
--- a/vendor/engines/saas/app/jobs/base_login_job.rb
+++ b/vendor/engines/saas/app/jobs/base_login_job.rb
@@ -1,0 +1,36 @@
+class LoginJob < ActiveJob::Base
+
+  def perform(params)
+    payload = map_payload(params)
+    Analytics::IdentifyUserJob.perform_later(payload)
+
+    #email = map_email(params)
+    #Queue job to send the email
+  end
+
+  def self.action(action)
+    @action = action
+  end
+
+  :private
+
+  def map_payload(params)
+    if params['emails'].is_a?(Array)
+      email = params['emails'].detect{|e| e['primary'] == true} || {}
+      params['user']['email'] = email['email']
+      params['user']['email_verified'] = email['verified']
+    end
+
+    params['user']['emails'] = params['emails']
+    params['user']['action'] = @action
+
+    {
+      'current_user' => params['user'],
+      'data' => params['user']
+    }
+  end
+
+  def map_email(params)
+    #some logic
+  end
+end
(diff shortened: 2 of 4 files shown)


vendor/engines/saas/lib/saas/engine.rb
```diff
@@ -17,6 +17,19 @@ class Engine < ::Rails::Engine
         })
       end
 
+      LoginController.class_eval do
+        before_action :page_job, only: [:public, :private]
+
+        :private
+
+        def page_job
+          if request.referer !~ /github\.com/
```
dahlbyk: This seems kind of brittle - what if we checked for one of the OAuth parameters?

dahlbyk:
This looks pretty good, will kick the tires on the review app after lunch and hopefully :ship: :it:.

vendor/engines/saas/lib/saas/engine.rb
discorick: I attempted to filter out by OAuth params `code` & `state`, but theres actually 3 calls on `/login/scope` during the process of authentication - 2 of which have no OAuth params which creates duplicate pages.

dahlbyk:
I'm seeing a triple-tap on `/login/{scope}` in Segment, all from Ruby.
image

dahlbyk:
> I'm seeing a triple-tap on `/login/{scope}` in Segment, all from Ruby.

To be more specific, it only seems to triple-tap if I've already authorized GitHub. From a clean/revoked state the event count is correct.

discorick:
Ah, ill address that good catch

discorick:
@dahlbyk , I can deal with the triple tap easily enough - having a double is a lot more difficult - I can't find any distinguishing request data between the initial login and the initial redirect. Any thoughts ?

dahlbyk:
> I can't find any distinguishing request data between the initial login and the initial redirect check.

Didn't quite parse that. Can you be more specific?

discorick:
In the situation where you are already authorized, on `/login/private` the page is called once, then the controller action executes up to `github_authenticate! :private` . Once warden is satisfied the token is good, the controller then redirects to `login/private` again. I am just looking for a differentiator for this second request to filter the page out

```ruby
 def private
    request.env['warden'].logout if github_authenticated? :default
    github_authenticate! :private # <<<< Right Here
    @user = gh.user
    @emails = @user.emails.all
    redirect_to params[:redirect_to] || "/"
  end
```

vendor/engines/saas/lib/saas/engine.rb
```diff
@@ -17,6 +17,20 @@ class Engine < ::Rails::Engine
...
       end
 
+      LoginController.class_eval do
+        before_action :page_job, only: [:public, :private]
+
+        :private
+
+        def page_job
+          auth_request = request.params['code'] && request.params['state']
+          if request.referer !~ /github\.com/ && !logged_in? && !auth_request
```
dahlbyk: Am I right in thinking the `request.referer` check is redundant given the other two?

1. `/login`
2. `/login/{scope}` (not `logged_in?` and `!auth_request`, so page is logged)
3. https://github.com/login/oauth/authorize
4. `/login/{scope}?code=...&state=...` (`auth_request` is true)
5. `/login/{scope}` (`logged_in?` should now be true)

discorick: True of theres already a token, referrer is needed for a login that is not yet authorized with GitHub, as the code and state are not yet available in request #2.

dahlbyk: Right, but this particularly code is mean to log `/login/{scope}` only at step 2. Step 5 handles the post-auth tracking of `/login/{scope}/authorized`, unless I'm missing something?

discorick: Yes, bouncing the referrer on an unauthenticated login prevents a step #4 duplicate Page (`state` and `code`) aren't part of the params just yet in this scenario.

dahlbyk: :OK: let's leave it. This is in the SaaS engine so Enterprise should not be affected, correct?

dahlbyk:
Am I missing something or did the latest here not get deployed?

vendor/engines/saas/lib/saas/engine.rb
discorick: Correct

discorick:
@dahlbyk Not sure what you mean ?

dahlbyk:
I don't see a deploy notification after the commit you pushed most recently.

dahlbyk:
> I don't see a deploy notification after the commit you pushed most recently.

Confirmed. 08e9657 was not deployed for some reason.
firefox_2016-03-12_11-27-08

Recreating Review App.

dahlbyk:
This is looking good.

One edge case we're not handling:

1. Revoke app
2. Login with public access
3. Hit `/repositories/private/{login}`
4. Approve extended permissions (N.B. pleased to see the `APP_NAME` redirect works in this scenario)
5. Auth succeeds without being tracked

This is probably somewhat rare, but if we wanted to test https://github.com/huboard/huboard-web/pull/181 we'd want to be able to tell when private auth succeeds. Any ideas?

discorick:
Another good catch @dahlbyk , looks like logins via `/repositories/(scope)/login` are handled via the `DashboardsController`

discorick:
@dahlbyk, just looked into this in more detail. 

I think for the time being it might be worth leaving this pathway un-tracked as `/login/private/authorized` - whats actually happening is a scope change, and complimentary to #181 I think we should release analytics surrounding the upgrading/downgrading of scopes in the improved login flow.

dahlbyk:
> I think we should release analytics surrounding the upgrading/downgrading of scopes in the improved login flow.

:+1: Merged as-is.

Also related to https://github.com/huboard/huboard-web/issues/172#issuecomment-196328505

rauhryan:
nice 👍

PR #233, opened 2016-03-11
Adds analytics for board creating

discorick:
Tracks if boards are being created after board create prompt


dahlbyk:
Simple enough. :+1:

PR #234, opened 2016-03-13
Add Vagrant .bashrc

dahlbyk:
My initial goal was an alias to simplify running the app (I picked `s` in honor of `rails s`, but am not attached to it).

As a means to that end, I introduced a `.bashrc` to be referenced from Vagrant. Should make iterating on that experience easier.

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -1,2 +1,12 @@
 #!/usr/bin/env bash
 cd /srv/huboard
+alias i="bundle install"
+alias s="bundle exec rails server --binding 0.0.0.0"
+
+echo "
+Welcome to Huboard!
+
+1. If you haven't already, run 'i' to install dependencies.
+2. Run 's' to start Rails.
+3. Browse to http://localhost:3001/ from your host.
+"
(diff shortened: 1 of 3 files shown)


discorick:
A more puppet-esq way to to accomplish this:

- Add `.bashrc` to `puppet/local_modules/bash/files`
- In the puppet manifest add the following:

```ruby
file { `/vagrant/.bashrc`:
  ensure => present,
  owner => 'vagrant',
  group =>   'vagrant'
  source => 'puppet://modules/bash/.bashrc`
} 
```

dahlbyk:
If I'm reading that correctly, won't that replace `.bashrc` altogether? (I assume you mean `file { '/home/vagrant/.bashrc`: ... }`?)

There's already a `.bashrc` with stuff in it, so I've been trying not to overwrite that.

discorick:
Yes you are right `file { '/home/vagrant/bashrc/: ...`

I would like to start putting together a source controlled `.bashrc` anyways, lets pick what we like from the image as well.

dahlbyk:
I think I'm going to try a different approach.

PR #237, opened 2016-03-15
Bump SaaS Rails to 4.2.4

dahlbyk:
I didn't realize we have 2 × `Gemfile.lock`. As we did with #217, pulling the SaaS Rails dependency version bump out of #211 to be deployed independently.

In addition to the Rails bump, I've synchronized a few other version mismatches relative to the root `Gemfile.lock`.


discorick:
Are changes in `1eadbaf` out of band changes to the Gemfile.lock?

dahlbyk:
> Are changes in 1eadbaf out of band changes to the Gemfile.lock?

Syncing versions changed elsewhere, yes (e.g. `json` version was bumped in https://github.com/huboard/huboard-web/commit/2cec97fbaefe21ed1f8e8836488f85ff0aa8564a).

Why do we even have a second Gemfile under SaaS?

discorick:
I wouldn't make any changes to the lock file - it will just rewrite itself as soon as anyone runs a 'bundle'. As a rule of thumb Gemfile.lock's should never be directly altered.

> Why do we even have a second Gemfile under SaaS?
Technically a Rails engine is a self-contained Rails application that is injected like a Gem.

dahlbyk:
Also, while we're fiddling with versions, I just pushed a tweak to bower.json to avoid https://github.com/emberjs/ember.js/pull/12787.

> I wouldn't make any changes to the lock file - it will just rewrite itself as soon as anyone runs a 'bundle'. As a rule of thumb Gemfile.lock's should never be directly altered.

FWIW, if I revert my Gemfile.lock changes (but leave the gemspec change), a `bundle install` reverts exactly to what I am committing here. It seems I got to it the wrong way, but arrived at correct results. :grinning:

discorick:
Ok haha then I support!

PR #238, opened 2016-03-15
Makes Slack integration avatars Enterprise friendly

discorick:
Derives the avatar url from the user payload. 

Closes https://github.com/huboard/huboard/issues/619
Fixes https://github.com/huboard/huboard/issues/614

### DIFF
--- a/lib/hu_board/services/slack.rb
+++ b/lib/hu_board/services/slack.rb
@@ -29,7 +29,7 @@ module HuBoard
     def transform_moved(mash)
       {
         username: "#{mash.meta.user.login} via HuBoard",
-        icon_url: "https://avatars.githubusercontent.com/u/#{mash.meta.user.id}?s=64",
+        icon_url: "#{mash.payload.issue.user.avatar_url}s=64",
         attachments: [
           {
             fallback: "<#{URI.join(github_url,mash.meta.user.login)}|@#{mash.meta.user.login}> moved <#{mash.payload.issue.html_url}|#{mash.meta.repo_full_name}##{mash.payload.issue.number}> from #{mash.payload.previous.text} to #{mash.payload.column.text }",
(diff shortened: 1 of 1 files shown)


lib/hu_board/services/slack.rb
```diff
@@ -29,7 +29,7 @@ def github_url
     def transform_moved(mash)
       {
         username: "#{mash.meta.user.login} via HuBoard",
-        icon_url: "https://avatars.githubusercontent.com/u/#{mash.meta.user.id}?s=64",
+        icon_url: "#{mash.payload.issue.user.avatar_url}s=64",
```
dahlbyk: Shouldn't there be a `?` or `&` before `s=64`?

discorick: An `&` correct, Faraday must have just parsed it in for me

dahlbyk: https://api.github.com/users/dahlbyk has `"avatar_url": "https://avatars.githubusercontent.com/u/133987?v=3"`, but I don't think we should assume the URL will always have a query string.

discorick: That's a good point, custom enterprise image servers might not be tacking API versioning onto their URL's. I didn't think of that

dahlbyk: Actually, let's just drop the size and let Slack browser/client scale appropriately.

firefox_2016-03-15_18-18-23

discorick: lol. Nice

lib/hu_board/services/slack.rb
```diff
@@ -29,7 +29,7 @@ def github_url
     def transform_moved(mash)
       {
         username: "#{mash.meta.user.login} via HuBoard",
-        icon_url: "https://avatars.githubusercontent.com/u/#{mash.meta.user.id}?s=64",
+        icon_url: mash.payload.issue.user.avatar_url,
```
dahlbyk: Didn't notice this before... Why `payload.issue.user` instead of `meta.user.login`? I would think they should match?

PR #239, opened 2016-03-15
Make HipChat integration enterprise friendly

### LINKED ISSUE #160: Allow URL for HipChat Integration
Currently the HipChat Integration is hard-coded, this should be an optional string to allow on-prem hipchat integrations to work as well.

discorick:
The api endpoint is now configurable for HipChat

Fixes #160

### DIFF
--- a/ember-app/app/templates/settings/integrations/hipchat.hbs
+++ b/ember-app/app/templates/settings/integrations/hipchat.hbs
@@ -9,6 +9,10 @@
     <label>
       {{input value=model.authToken placeholder="Authorization Token" required=true }}
     </label>
+    <p> HipChat Endpoint (no trailing slashes): </p>
+    <label>
+      {{input value=model.endpoint required=true }}
+    </label>
     <button disabled={{disabled}} class="hb-button">Submit webhook</button>
     <div class="with-line"><span>or</span></div>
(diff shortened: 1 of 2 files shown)


ember-app/app/templates/settings/integrations/hipchat.hbs
```diff
@@ -9,6 +9,10 @@
     <label>
       {{input value=model.authToken placeholder="Authorization Token" required=true }}
     </label>
+    <p> HipChat Endpoint (no trailing slashes): </p>
+    <label>
+      {{input value=model.endpoint required=true }}
```
drusellers: Is it worth having some example values?

discorick: The value is pre filled as the default HipChat API endpoint `"https://api.hipchat.com"` as per `ember-app/app/controllers/settings/integrations.js` Line 8, are you suggesting even more examples ?

dahlbyk:
Change looks good. Do we care to drop a trailing slash if the user happens to specify one (resulting in `https://api.hipchat.com//v2/room/foo/notification?auth_token=bar`?

dahlbyk:
Oh.

> HipChat Endpoint (no trailing slashes):

PR #240, opened 2016-03-16
Consistent Word Break Handling

### LINKED ISSUE #205: Consistent Word Break Handling
We have implemented variations on this theme in several places: #124 #198  #199 #202. None of them seem to be working as expected in Firefox:

https://huboard.com/dotnet/roslyn columns aren't equal width, and on https://huboard.com/dotnet/roslyn#/issues/56883993:
firefox_2016-02-17_16-43-15

So two things:

1. We need to stop using [`word-break: break-all`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) altogether. `word-break: break-word` seems fine but is Chrome-specific.
2. Firefox does support [`word-wrap: break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap) (an old property since renamed to [`overflow-wrap`](https://docs.webplatform.org/wiki/css/properties/overflow-wrap) - we should probably include both), but there are caveats to it working as we need it to within `display: flex`: https://bugzilla.mozilla.org/show_bug.cgi?id=1136818.

We need a standard pattern for this to be implemented everywhere we render user-generated text.

dahlbyk:
Closes #205 using a global `word-wrap: break-word` rule plus a few `min-width:0` hacks due to Firefox ([supposedly spec-defined](https://bugzilla.mozilla.org/show_bug.cgi?id=1136818#c11)) interaction between `word-wrap` and Flexbox.

### DIFF
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -302,8 +302,6 @@
     color: $darkGrey;
   }
   .title {
-    word-wrap: break-word;
-    @include word-break(break-word);
     white-space: normal;
     text-overflow: clip;
     overflow:auto;
--- a/app/assets/stylesheets/_issue.scss
+++ b/app/assets/stylesheets/_issue.scss
@@ -57,7 +57,6 @@
 }
 
 .flex-issue-title {
-  @include word-break(break-word);
   @include display(flex);
   .issue-closed-badge p {
     margin: 4px 4px;
--- a/app/assets/stylesheets/_mixins.scss
+++ b/app/assets/stylesheets/_mixins.scss
@@ -91,7 +91,3 @@
     transform: translate(0, 2px);
   }
 }
-
-@mixin word-break($value) {
-  @include prefixer(word-break, $value, webkit moz ms spec);
-}
--- a/app/assets/stylesheets/_overlay.scss
+++ b/app/assets/stylesheets/_overlay.scss
@@ -154,15 +154,6 @@ body > .fullscreen-overlay {
 }
 .fullscreen-card-description {
   p {
-  -ms-word-break: break-all;
-     word-break: break-all;
-
-     // Non standard for webkit
-     word-break: break-word;
-
--webkit-hyphens: auto;
-   -moz-hyphens: auto;
-        hyphens: auto;
   }
 }
 .fullscreen-card-right{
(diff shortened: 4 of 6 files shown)


dahlbyk:
The current implementation does seem to resolve all known word-wrapping touch points, however it introduces a change in task board column sizing. The new sizing is consistent between Chrome and Firefox, but introduces wider columns and a (hidden) horizontal scroll rather than splitting within available space. Any Flexbox wizards know how to fix this?


Basically the Task columns now behave like Milestone columns, which are also handled consistently:


And for reference, issue modal title and description work too (actually an improvement in Chrome - check the issue number):

dahlbyk:
https://huboard-rails-pr-240.herokuapp.com/rauhryan/skipping_stones_repo
https://huboard-rails-pr-240.herokuapp.com/rauhryan/skipping_stones_repo#/milestones

Open: https://huboard-rails-pr-240.herokuapp.com/rauhryan/skipping_stones_repo#/issues/132264941
Closed: https://huboard-rails-pr-240.herokuapp.com/rauhryan/skipping_stones_repo#/issues/140275058

discorick:
Looking good, as you mentioned there is work to do with the column widths. An example would be resizing the window, the first column "reaches" towards the right hand side of the viewport at a certain size:

image

dahlbyk:
> An example would be resizing the window, the first column "reaches" towards the right hand side of the viewport at a certain size

That's actually an existing behavior from `@media` queries: https://github.com/huboard/huboard-web/blob/4a6febb6015dfaa620ac8ed3b48100b71ca87edb/app/assets/stylesheets/flex_layout.css.scss#L382-L422

discorick:
Oh I see, interesting, I suppose thats a different UX discussion altogether.

dahlbyk:
> I suppose thats a different UX discussion altogether.

Si

rauhryan:
rauhryan_skipping_stones_repo_huboard_and_rauhryan_skipping_stones_repo_huboard_and_itunes

Looks like setting a max-width on the containing div is all that was needed.

rauhryan:
Link for review: https://huboard-rails-pr-240.herokuapp.com/rauhryan/skipping_stones_repo

dahlbyk:
> Looks like setting a max-width on the containing div is all that was needed.

I like simple solutions.

This checks out for me in Firefox/Chrome on Windows, for Tasks and Milestones. :shipit:?

PR #241, opened 2016-03-16
Falls back to an avatar on S3 for enterprise slack avatars

discorick:
Addresses comments in https://github.com/huboard/huboard/issues/619

### DIFF
--- a/lib/hu_board/services/slack.rb
+++ b/lib/hu_board/services/slack.rb
@@ -26,10 +26,17 @@ module HuBoard
       ENV["GITHUB_WEB_ENDPOINT"] || "https://github.com/"
     end
 
+    def avatar_url(mash)
+      if github_url != 'https://github.com/'
+        return 'https://s3-us-west-2.amazonaws.com/prod.enterprise.huboard.com/assets/slack/avatar.png'
+      end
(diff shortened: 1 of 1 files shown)


lib/hu_board/services/slack.rb
```diff
@@ -26,10 +26,17 @@ def github_url
       ENV["GITHUB_WEB_ENDPOINT"] || "https://github.com/"
     end
 
+    def avatar_url(mash)
+      if github_url != 'https://github.com/'
```
discorick: @rauhryan @dahlbyk this is the best way I know to differentiate between enterprise and not on the fly, if theres a better key let me know.

dahlbyk: I assumed there would be an environment variable we could check...?

dahlbyk:
Change looks good other than the question of identifying enterprise. I'm not a fan of the URL check, so if there isn't a clean way to identify Enterprise let's add one. (Even if that just means committing some sort of global `enterprise = true` change on the enterprise branch.)

discorick:
Cool, I just need @rauhryan to glance at the ENV var and let me know if theres something better in place already.

PR #243, opened 2016-03-16
Proposing an issue template for huboard-web

discorick:
Let tighten up our new issue creation! Looking for suggestions and feedback.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,7 @@
+#### Description:
+
+#### Steps to reproduce (if applicable):
+
+#### Helpful external resources (blog, docs etc.):
+
+#### Notes:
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+#### Description:
+
+#### Steps to reproduce (if applicable):
+
+#### Helpful external resources (blog, docs etc.):
```
dahlbyk: Comma after docs :trollface:

dahlbyk:
Do we want to give any guidance as to what might be opened here vs on huboard/huboard? (I honestly don't know what our guidance would be; I've mostly lived here.)

discorick:
I think now is a good time to have that conversation - theirs a few ways we could go about doing it, like locking issues in one way or another. I think @rauhryan was thinking bugs should be opened on huboard-web and feature requests via huboard/huboard

dahlbyk:
> I think @rauhryan was thinking bugs should be opened on huboard-web and feature requests via huboard/huboard

I'd be fine trying to funnel everything external through huboard/huboard, reserving repo-specific issues for bugs/debt that we identify. We might even make the template here simply a comment referring people to https://github.com/huboard/huboard-web/issues/new.

PR #244, opened 2016-03-17
Increases the history limit for charges

discorick:
Currently the history limit for charges is 10, for tax purposes are larger limit is sometimes required, bumping this up to 20.


dahlbyk:
Confirmed `limit` in the API: https://stripe.com/docs/api?lang=ruby#list_charges-limit

PR #247, opened 2016-03-21
Routes 503's and 400's to 'server_error'

discorick:
When the GitHub API goes down currently users are greeting with ugly plain-text pages - this PR routes them to the server-error endpoint.


dahlbyk:
:cool:

PR #248, opened 2016-03-23
Enables default issue templates on HuBoard

### LINKED ISSUE #206: Default Templates for New issues
Give HuBoard the ability to use new issue templates as per https://github.com/dear-github/dear-github/issues/125

Official docs: https://help.github.com/articles/creating-an-issue-template-for-your-repository/
Announcement: https://github.com/blog/2111-issue-and-pull-request-templates

discorick:
Closes #206

### DIFF
--- a/app/controllers/api/board_controller.rb
+++ b/app/controllers/api/board_controller.rb
@@ -12,7 +12,7 @@ module Api
     def issue_template
       board = huboard.board(params[:user], params[:repo])
       template = board.issue_template
-      decoded_template = Base64.decode64(template['content'])
+      decoded_template = template ? Base64.decode64(template['content']) : nil
       render json: { issue_template: decoded_template }
     end
   end
--- a/app/controllers/api/board_controller.rb
+++ b/app/controllers/api/board_controller.rb
@@ -11,9 +11,8 @@ module Api
 
     def issue_template
       board = huboard.board(params[:user], params[:repo])
-      template = board.issue_template
-      decoded_template = template ? Base64.decode64(template['content']) : nil
-      render json: { issue_template: decoded_template }
+      template_content = board.issue_template_content
+      render json: { issue_template: template_content }
     end
   end
 end
--- a/app/controllers/api/board_controller.rb
+++ b/app/controllers/api/board_controller.rb
@@ -9,5 +9,11 @@ module Api
       render json: huboard.board(params[:user], params[:repo]).commit(params[:commit])
     end
 
+    def issue_template
+      board = huboard.board(params[:user], params[:repo])
+      template = board.issue_template
+      decoded_template = Base64.decode64(template['content'])
+      render json: { issue_template: decoded_template }
+    end
   end
 end
--- a/ember-app/app/controllers/issue/create.js
+++ b/ember-app/app/controllers/issue/create.js
@@ -8,7 +8,11 @@ var IssuesCreateController = Ember.Controller.extend({
     return this.get('allRepos.firstObject');
   }.property('allRepos.[]'),
   repoChanged: function(){
-    this.set('model.body', this.get('selectedRepo.issue_template'));
+    var template = this.get('selectedRepo.issue_template');
+    var previous_template = this.get('previousRepo.issue_template');
+    if(this.get('issueBodyDirty') && !template){ return; }
+    if(previous_template && this.get('model.body') !== previous_template){ return; }
+    this.set('model.body', template);
   }.observes('selectedRepo'),
   otherLabels: Ember.computed.alias("selectedRepo.data.other_labels"),
   assignees: Ember.computed.alias("selectedRepo.data.assignees"),
@@ -21,11 +25,14 @@ var IssuesCreateController = Ember.Controller.extend({
     return this.get("model.title");
   }.property("model.title"),
   issueBodyDirty: function(){
-    if(this.get('selectedRepo.issue_template')){
-      return this.get('model.body') !== this.get('selectedRepo.issue_template');
+    if(this.get('previousRepo.issue_template')){
+      return this.get('model.body') !== this.get('previousRepo.issue_template');
     }
     return !Ember.isEmpty(this.get('model.body'));
   }.property('selectedRepo.data.repo.name', 'model.body'),
+  promptForRepoChange: function(repo){
+    return this.get('issueBodyDirty');
+  },
   mentions: function(){
     return _.uniq(_.compact(this.get('selectedRepo.assignees')), function(i){
       return i.login;
@@ -44,8 +51,8 @@ var IssuesCreateController = Ember.Controller.extend({
         controller.get("target").send("closeModal");
       });
     },
-    assignRepo: function(repo, callback){
-      if(!this.confirmChange(callback)){ return };
+    assignRepo: function(repo){
+      this.set('previousRepo', this.get('selectedRepo'));
       var get = Ember.get;
 
       // transfer assignee if possible
@@ -64,14 +71,6 @@ var IssuesCreateController = Ember.Controller.extend({
       });
       this.set("model.labels", commonLabels);
     }
-  },
-  confirmChange: function(callback){
-    var confirmation = true;
-    if(this.get('issueBodyDirty')){
-      var confirmation = confirm('Any changes to the current issue will be lost!')
-    }
-    callback(confirmation);
-    return confirmation;
   }
 });
 
--- a/ember-app/app/controllers/issue/create.js
+++ b/ember-app/app/controllers/issue/create.js
@@ -20,6 +20,12 @@ var IssuesCreateController = Ember.Controller.extend({
   isValid: function () {
     return this.get("model.title");
   }.property("model.title"),
+  issueBodyDirty: function(){
+    if(this.get('selectedRepo.issue_template')){
+      return this.get('model.body') !== this.get('selectedRepo.issue_template');
+    }
+    return !Ember.isEmpty(this.get('model.body'));
+  }.property('selectedRepo.data.repo.name', 'model.body'),
   mentions: function(){
     return _.uniq(_.compact(this.get('selectedRepo.assignees')), function(i){
       return i.login;
@@ -38,7 +44,8 @@ var IssuesCreateController = Ember.Controller.extend({
         controller.get("target").send("closeModal");
       });
     },
-    assignRepo: function(repo){
+    assignRepo: function(repo, callback){
+      if(!this.confirmChange(callback)){ return };
       var get = Ember.get;
 
       // transfer assignee if possible
@@ -57,6 +64,14 @@ var IssuesCreateController = Ember.Controller.extend({
       });
       this.set("model.labels", commonLabels);
     }
+  },
+  confirmChange: function(callback){
+    var confirmation = true;
+    if(this.get('issueBodyDirty')){
+      var confirmation = confirm('Any changes to the current issue will be lost!')
+    }
+    callback(confirmation);
+    return confirmation;
   }
 });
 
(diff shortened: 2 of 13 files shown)


test/lib/bridge/content_test.rb
```diff
@@ -0,0 +1,100 @@
...
+      it 'looks for a template in the root of the repo' do
+        @mock_gh_instance.stubs(:contents).returns(file_list)
+        @mock_gh_instance.expects(:contents).with('').once
+        @mock_gh_instance.expects(:contents).with('ISSUE_TEMPLATE.txt').returns(file_list[0])
+
+        assert_equal file_list[0], bridge.issue_template
+      end
+
+      it 'looks for a template in the .github directory' do
+        @mock_gh_instance.expects(:contents).with('').returns []
```
dahlbyk: It occurs to me that this would include at least the `.github` folder, which we could check for in the root response before making the directory listing.

test/lib/bridge/content_test.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      it 'it returns nil' do
+        @mock_gh_instance.expects(:contents).with('').returns file_list
+
+        assert_equal nil, bridge.issue_template
+      end
+    end
+
+    describe 'the repository is empty' do
+      file_list = ["message", "This repository is empty."]
```
dahlbyk: Is the empty-repo response really an array? I would expect a hash.

test/lib/bridge/content_test.rb
dahlbyk: Also, you might consider `it "finds a template in the .github directory"` in contrast with a new `it "does not find a template in the .github directory"` test case.

lib/bridge/github/repo.rb
```diff
@@ -55,13 +56,17 @@ def details(data={})
...
+
       {
         columns: columns,
         milestones: milestones,
         other_labels: other_labels.sort_by {|l| l['name'].downcase },
         link_labels: link_labels,
         assignees: assignees,
-        issues: issues
+        issues: issues,
+        issue_template: decoded_template
```
dahlbyk: Does Ruby support trailing commas on hashes and such? If so, I'd suggest we get in the habit of including them simply because it cleans up future diffs.

lib/bridge/github/contents.rb
```diff
@@ -0,0 +1,20 @@
+class Huboard
+  module Contents
+    ISSUE_TEMPLATE_PATTERN = /^ISSUE_TEMPLATE\.(md|txt)/i
+
+    def issue_template
+      file_list = gh.contents ''
+      return find_template(file_list) if file_list && file_list.size > 0
+
+      file_list = gh.contents '.github'
```
dahlbyk: As I mentioned below, we should be able to check if `.github` exists in `file_list` before making this API call.

lib/bridge/github/contents.rb
```diff
@@ -0,0 +1,20 @@
+class Huboard
+  module Contents
+    ISSUE_TEMPLATE_PATTERN = /^ISSUE_TEMPLATE\.(md|txt)/i
```
dahlbyk: Add `$` at the end of the pattern? This would match `issue_template.md5`.

lib/bridge/github/repo.rb
seanlinsley: yes it does

lib/bridge/github/contents.rb
discorick: :+1:

test/lib/bridge/content_test.rb
discorick: Yep, I am not sure if that's a Ghee thing, or the GitHub API

app/controllers/api/board_controller.rb
```diff
@@ -9,5 +9,11 @@ def commit
       render json: huboard.board(params[:user], params[:repo]).commit(params[:commit])
     end
 
+    def issue_template
+      board = huboard.board(params[:user], params[:repo])
+      template = board.issue_template
+      decoded_template = template ? Base64.decode64(template['content']) : nil
```
dahlbyk: I'm surprised to have a `Base64.decode64` in two places. Could the decode move down into bridge?

dahlbyk:
I assume you've been testing this with real repos - links? At some point we might want to set up an org in which to set up "known state" repos like this.

Scenarios off the top of my head:

- Repo with nothing
- Repo with no root template and empty `.github/`
- Repo with no root template and no `.github/` template
- Repo with root template and no `.github/` template
- Repo with no root template and a `.github/` template
- Repo with enough root files to push template to second page in API (does Ghee handle?)

Multiple template options (do we align with GH in picking "winner"?):

- Repo with root template and `.github/` template
- Multiple root templates
- Multiple `.github/` templates

discorick:
> I assume you've been testing this with real repos - links?

In my acceptance testing I have been running scenarios against real private repos of mine. The test suites in the app are just unit tests, just mocking out the Ghee behaviour (which is pretty predictable). 

> At some point we might want to set up an org in which to set up "known state" repos

We do have a test org at https://github.com/huboard-testing where we could start populating these kind of repos. I agree that would be a good idea 

> do we alight with GH in picking "winner"?

The [docs](https://help.github.com/articles/creating-an-issue-template-for-your-repository/) don't even specify who that is, I am happy with `root` > `.github`

dahlbyk:
> do we alight with GH in picking "winner"?
> 
> The docs don't even specify who that is, I am happy with `root` > `.github`

I'm fine with that too, but we could confirm by setting up a test repo with both (or all four, really). This is certainly an edge case...maybe it's not even worth testing for consistency as long as we show _a_ template?

GitHub really just needs to provide an API to identify *the* file.

dahlbyk:
Revisions so far look good.

discorick:
Thanks just playing around with detecting whether `.github` is listed

dahlbyk:
`.github` check looks correct. Can you add a test to confirm that we skip the contents check if the directory doesn't exist?

Can you give me access to huboard-testing so I can set up some template examples?

discorick:
>  Can you add a test to confirm that we skip the contents check if the directory doesn't exist?

Yes! Sorry I meant to do that in the last commit, slipped my mind as I got fixated on something else.

test/lib/bridge/content_test.rb
```diff
@@ -26,6 +26,7 @@ class MockBridge
       it 'looks for a template in the root of the repo' do
         @mock_gh_instance.expects(:contents).with('').returns file_list
         @mock_gh_instance.expects(:contents).with('ISSUE_TEMPLATE.txt').returns(file_list[0])
+        @mock_gh_instance.expects(:contents).with('.github').never
```
dahlbyk: :+1:

dahlbyk:
Added various test repos, all linked from https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-links (which doesn't seem to work - not sure if I messed up a link?).

* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/empty
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/no-files
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-github-both
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-github-md
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-github-txt
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-none-with-github
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-none-without-github
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-root-and-github-both
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-root-both
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-root-md
* https://huboard-rails-pr-248.herokuapp.com/huboard-testing/issue-templates-root-txt

dahlbyk:
Two observations:

1. https://github.com/huboard-testing/issue-templates-root-and-github-both/issues/new does confirm that GitHub prefers `.github/` over root (we're wrong), and `.md` over `.txt` (we're right). An edge case that's resolved easily enough ("Don't have two templates, silly."), but I thought I'd point it out for completeness.

2. Template switching in linked repos is working great, but changing repos after I begin filling in a template replaces my glorious message with a template again. We should probably suppress the template swap (with an offer to use the new template?) if the description has been edited.

-------------------------------------------------------------------

> which doesn't seem to work

Apparently fixed by last commit. As shared offline:

> It was Ghee being a weirdo. The contents call wasn’t failing tests when it should have cause of how the ghee proxy is handling some cases.

The array check is fine, but it would probably be more correct for Ghee to check the status code.

- https://api.github.com/repos/huboard-testing/no-files/contents returns 200 with `[]`
- https://api.github.com/repos/huboard-testing/empty/contents returns 404 with the error message

dahlbyk:
> Adds a confirmation alert notifying that repo changes will wipe the current issue body

Hrm....I'm generally anti-`alert`. I know this isn't the only use thereof, so if I defer that objection here are two ideas that should make for fewer prompts:

1. Always allow changing to a repo without a template
   - Currently you can't even switch between two template-less repos if the description has been edited.
   - Beyond that, I don't see a problem with continuing to use another repo's template if you switch to a repo that doesn't give issue guidance
2. Allow changing between repos with templates if the current description is empty
   - Select All :point_right: Cut :point_right: change repo :point_right: Select All :point_right: Paste

That being said, rather than block the repo change due to the template conflict, what if we:

1. Allowed the repo change (that's what I meant by "suppress the template swap")
2. (optional) Non-intrusively warned that the new repo has a different template
3. (optional) Provided an Apply Template option that would overwrite with the new template

discorick:
> 1. Always allow changing to a repo without a template
>   - Currently you can't even switch between two template-less repos if the description has been edited.
>   - Beyond that, I don't see a problem with continuing to use another repo's template if you switch to a repo that doesn't give issue guidance

re: 1a I am honestly not sure what our users would prefer, a naggy alert, or loosing whatever work is in there. Alerts you can tell to dismiss next time, which seems better to me as a matter of personal taste

re 1b Seems like a great idea to me

> 2. Allow changing between repos with templates if the current description is empty
>   - Select All :point_right: Cut :point_right: change repo :point_right: Select All :point_right: Paste

A lot like '1b', Also seems like a reasonable UX to me

For the remaining points, I would rather prompt with an updated alert (something like "changing templates will clear any current work") until we can have a proper design laid out for this. Again this is one of those places where I don't know what our users would prefer, or what they expect to happen. 

I am open to the idea of just swapping between templates with non-intrusive warning I suppose - but I wouldn't want to spend any design time making a fancy warning. Maybe one of our current toasts would do at the top of the screen?

dahlbyk:
My vote is to skip any new UX for now, alert or otherwise:

1. If description is empty or matches previous template, apply new template
2. Otherwise, leave modified description alone.

discorick:
Yeah, I agree, I was just walking down the alerts as you commented.

dahlbyk:
Reading comprehension fail, sorry.

> ...but I wouldn't want to spend any design time making a fancy warning. Maybe one of our current toasts would do at the top of the screen?

Meh, let's see what feedback we get without any extra UX effort.

dahlbyk:
> Transfers new issue descriptions more intelligently

I'm still losing an edited description sometimes.

template-quirks

ember-app/app/controllers/issue/create.js
```diff
@@ -17,6 +24,15 @@ var IssuesCreateController = Ember.Controller.extend({
   isValid: function () {
     return this.get("model.title");
   }.property("model.title"),
+  issueBodyDirty: function(){
+    if(this.get('previousRepo.issue_template')){
+      return this.get('model.body') !== this.get('previousRepo.issue_template');
+    }
+    return !Ember.isEmpty(this.get('model.body'));
```
dahlbyk: You might flip this check to be first - if the body is empty, we should never consider it dirty.

dahlbyk:
:sparkles: :clap: :sparkles: :clap: :sparkles: :clap: :sparkles: :clap: :sparkles: :clap: :sparkles:

PR #249, opened 2016-03-23
re-enables the copy board button if the board being copied doesn't have any columns

rauhryan:


discorick:
Nice! Good call on this one.

PR #257, opened 2016-04-01
Internet Explorer Fixes

dahlbyk:
Some fixes toward https://github.com/huboard/huboard/issues/20 https://github.com/huboard/huboard/issues/640

Hoping @JBertling can give me a hand here; my IE-fu is limited.

### DIFF
--- a/ember-app/app/initializers/sockets.js
+++ b/ember-app/app/initializers/sockets.js
@@ -72,9 +72,7 @@ export function initialize(container, application){
         return channel.replace(/\./g, '!').toLowerCase();
       },
       subscribeTo: function(channel) {
-        const channel = this._sanitizeChannel(channel);
-
-        self = this;
+        const channel = this._sanitizeChannel(channel), self = this;
         var client = this.get('client'), 
         callbacks = Ember.$.Callbacks();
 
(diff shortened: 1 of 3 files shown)


ember-app/app/initializers/sockets.js
```diff
@@ -72,9 +72,7 @@ export function initialize(container, application){
         return channel.replace(/\./g, '!').toLowerCase();
       },
       subscribeTo: function(channel) {
-        const channel = this._sanitizeChannel(channel);
-
-        self = this;
+        const channel = this._sanitizeChannel(channel), self = this;
```
discorick: Whoa interesting, was self undefined in IE11 ?

dahlbyk: > Variable undefined in strict mode

dahlbyk:
> Allow Ctrl-Enter (standard on Windows, vs Meta-Enter)

This is confirmed to work for commenting. I thought I remembered someone saying Meta-Enter should also open the Detail screen when creating issues, too, but that's not working for me.

dahlbyk:
I'm kind of tempted to :ship: this as-is to at least make IE not completely broken sooner rather than later. Can revisit with more polish at a later date. Thoughts?

discorick:
Looks good to me!

PR #258, opened 2016-04-03
Actually suppress publishing same-column milestone moves

dahlbyk:
Fixes https://github.com/huboard/huboard/issues/641. We're already trying to guard against this, but we're not using `before_perform` correctly. Beyond that, we don't actually want to prevent the job from firing altogether because that would also prevent card relocations from publishing to the client.


dahlbyk:
> Set/clear milestone after sending action 			fd3da98

@discorick noticed that milestone change notifications broke from the issue detail screen. Turns out they weren't being marked with `changed_milestones = true`. We need to set the selected milestone after sending the event that compares old and new.

PR #262, opened 2016-04-08
Sidebar Overflow / Remove App Footer?

dahlbyk:
I started out trying to fix filter overflow and double-scroll in IE:
iexplore_2016-04-08_11-25-44

But it turns out I've also made the header and footer sticky in Chome/Firefox:

It's still not working right in Milestones, but before I continue down the rabbit hole I thought I'd check if this is actually the preferred UX. My gut feeling is that a sticky header+footer would be cool, but the footer should be shorter. Thoughts?

Closes https://github.com/huboard/huboard/issues/640


app/assets/stylesheets/_board.scss
```diff
@@ -103,6 +103,7 @@
 
 .column h3 {
   line-height: 20px;
+  min-height: 42px;
```
dahlbyk: This feels wrong, but without it the header loses its bottom padding if the column is full....unsure why.
chrome_2016-04-08_10-43-52


/cc @JBertling

app/assets/stylesheets/flex_layout.css.scss
```diff
@@ -125,6 +126,7 @@ body {
     }
   }
    height: 40px;
+   min-height: 40px;
```
dahlbyk: This also feels wrong, but `height` alone is ignored:
chrome_2016-04-08_12-00-15

discorick:
Sticky footer would probably end up being a pretty unpopular move, real-estate is a big deal

dahlbyk:
> Sticky footer would probably end up being a pretty unpopular move, real-estate is a big deal

That's fair. What if we just dropped the footer altogether from the board view? The same links are available from the dashboard when authenticated.

discorick:
> What if we just dropped the footer altogether from the board view

I support this

dahlbyk:
> > What if we just dropped the footer altogether from the board view
> 
> I support this

I've swapped out the header menu icon for a logo:
iexplore_2016-04-18_01-57-20

And dropped the app footer:





So far, this PR does not change Firefox scroll behavior (i.e. it matches old FF/Chrome behavior), but does align IE/Chrome in having a sticky header.

dahlbyk:
Updated more explicit `overflow` rules and Firefox has fallen in line:
image

Milestones view is still not behaving. Any comments before I start on that?

discorick:
Im not digging the HB logo on the top left (might just be personal Pref)... I think we are planning on using that hamburger intelligently once the dashboard goes under way.

dahlbyk:
> Im not digging the HB logo on the top left (might just be personal Pref)... I think we are planning on using that hamburger intelligently once the dashboard goes under way.

Could we revisit the header/hamburger as part of designing the new dashboard? For now, the icon doesn't really align with its function. I also considered replacing with a home icon, but it seems wise to have at least _some_ sort of branding in the app itself. If not the full logo (literally copied from the footer), maybe the white-on-purple H?

dahlbyk:
Superseded by #280

dahlbyk:
I've added proof-of-concept styles for a fixed column header. Would need to redesign the column header styles, since we won't know the correct width for the `fixed` elements, and I'm really not if/how it would be possible to deal with column names that are too wide...

PR #263, opened 2016-04-08
Enable Optimizely in HuBoard SaaS

petey-rock:
Enable Optimizely in HuBoard SaaS

### DIFF
--- a/app/views/errors/not_found.html.erb
+++ b/app/views/errors/not_found.html.erb
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
+  <%= render partial: 'shared/abtest' %>
   <meta charset="utf-8" />
   <meta name="description" content="Instant project management for GitHub issues">
   <title>HuBoard :: Not found</title>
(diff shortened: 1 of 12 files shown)


dahlbyk:
We'll need to reference the partial in the other interesting views as well.

dahlbyk:
The `OPTIMIZELY_KEY` config value didn't come over without adding it to `app.json`, but I'm OK with that - we usually don't need it. I've fixed for just this review app, and removed from the parent:

```
heroku config:set OPTIMIZELY_KEY=5531021703 -a huboard-rails-pr-263
heroku config:remove OPTIMIZELY_KEY -a huboard-rails
```

Optimizely request is coming through as expected:
firefox_2016-04-09_08-57-38

dahlbyk:
It occurs to me that I may have merged this prematurely - I've restored the config value on huboard-rails if you want to test end-to-end in Optimizely before this hits production. (I'm pretty confident this will work just fine.)

discorick:
Ill look at it today!

PR #265, opened 2016-04-11
disables decoding attempts on nil issue templates

discorick:


lib/bridge/github/contents.rb
```diff
@@ -16,7 +16,8 @@ def issue_template
 
     def issue_template_content
       begin
-        template = Base64.decode64(issue_template['content'])
+        template = issue_template
+        template = Base64.decode64(template['content']) if template
```
dahlbyk: Not a fan of reusing `template` for a completely different thing. Suggest either:

1. Rename second use to `content`
2. Immediately `return` instead

Thoughts?

lib/bridge/github/contents.rb
```diff
@@ -16,13 +16,14 @@ def issue_template
 
     def issue_template_content
       begin
-        template = Base64.decode64(issue_template['content'])
+        template = issue_template
+        content = Base64.decode64(template['content']) if template
       rescue => e
-        ::Raygun.track_exception(e, custom_data: {error_response: template.to_s})
-        template = nil
+        ::Raygun.track_exception(e, custom_data: {error_response: content.to_s})
```
dahlbyk: This seems like it should stay `template`? And probably guard against `nil`, too.

PR #266, opened 2016-04-11
Force encodes issue template content to UTF-8

discorick:
Issue templates decoded on the back end sometime end up with ASCII characters that cause Rails to throw.

### DIFF
--- a/lib/bridge/github/contents.rb
+++ b/lib/bridge/github/contents.rb
@@ -23,7 +23,7 @@ class Huboard
         content = nil
       end
 
-      return content
+      return content ? content.force_encoding("utf-8") : content
     end
 
     :private
--- a/test/lib/bridge/content_test.rb
+++ b/test/lib/bridge/content_test.rb
@@ -106,7 +106,7 @@ class Huboard::ContentsTest < ActiveSupport::TestCase
       end
     end
 
-    describe 'The content string is not UTF-8' do
+    describe 'The content string decodes to ascii' do
       it 'force encodes an ascii string' do
         ascii= 'IyBVc2VyIHN0b3J5CgoqKkNvbW8qKgoqKmV1IHF1ZXJvKiogCioqcGFyYSBx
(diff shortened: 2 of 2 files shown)


lib/bridge/github/contents.rb
```diff
@@ -23,7 +23,7 @@ def issue_template_content
         content = nil
       end
 
-      return content
+      return content ? content.force_encoding("utf-8") : content
```
dahlbyk: Is there a reason to use `content ? content.force_encoding("utf-8") : content` over `content.force_encoding("utf-8") if content`?

test/lib/bridge/content_test.rb
```diff
@@ -105,6 +105,35 @@ class MockBridge
         assert_equal nil, bridge.issue_template_content
       end
     end
+
+    describe 'The content string is not UTF-8' do
+      it 'force encodes an ascii string' do
```
dahlbyk: I think technically the problem is that the content string _is_ UTF-8, but Ruby is decoding it as ASCII by default to we have to reencode as UTF-8. http://stackoverflow.com/a/16596591 describes what we're seeing pretty well.

lib/bridge/github/contents.rb
discorick: Ternary provides an explicit return value

dahlbyk: So the difference is that `""` would be switched to `nil` using the latter?

Doesn't really matter, just learning what idioms work when.

test/lib/bridge/content_test.rb
discorick: Interesting, yeah I see what you mean

lib/bridge/github/contents.rb
discorick: Its ensuring explicit `nil`, though its almost certain that a non-true content value would return `nil` anyways I prefer explicit over implicit in this kind of situation.

PR #268, opened 2016-04-13
adds a simple check to for Faye to prevent the white screen of death

rauhryan:


discorick:
What does https://github.com/huboard/huboard-web/pull/268/files#diff-5bb7f4bc8d062dea46291a11be6e7811R9 do ?

rauhryan:
@discorick is a null object with the same signature as the real socket object

So the issue controller doesn't throw an exception if the socket isn't configured

https://github.com/huboard/huboard-web/blob/master/ember-app/app/controllers/issue.js#L104

ember-app/app/initializers/sockets.js
```diff
@@ -6,12 +6,13 @@ export function initialize(container, application){
...
     correlationId : correlationId,
     sockets: {},
+    publish: Ember.K,
     subscribe: Ember.K,
     unsubscribe: Ember.K,
     subscribeTo: Ember.K
   });
 
-  if(application.get("socketBackend")){
+  if(application.get("socketBackend") & window.Faye){
```
discorick: Single `&`?

ember-app/app/initializers/sockets.js
```diff
@@ -6,12 +6,13 @@ export function initialize(container, application){
...
     correlationId : correlationId,
     sockets: {},
+    publish: Ember.K,
     subscribe: Ember.K,
     unsubscribe: Ember.K,
     subscribeTo: Ember.K
   });
 
-  if(application.get("socketBackend")){
+  if(application.get("socketBackend") &&  window.Faye){
```
dahlbyk: Would it be more durable to check if `typeof faye === 'function'`, since that's specifically what we need? Or should the `window.Faye` check live in `faye()`?

dahlbyk: Yeah, I'd check for `Faye` before we use it [here](https://github.com/huboard/huboard-web/blob/47242aba93d1b7641e599ca57e0f69e55def8b77/ember-app/app/utilities/messaging/fayeClient.js#L5).

PR #269, opened 2016-04-14
Fixes the broken 'issueUnlabeled' event

discorick:
Due to a type, `issue_unlabeled` events were failing on the client. Furthermore the exceptions generated by the failure prevents all further events to dispatch until refresh :( 

All of our users who have struggled through this deserve an apology. I am so sorry!

Additional bug I found: 
- The toast will never work in the current configuration (the timestamp check should always fail), I will address this in a future PR

### DIFF
--- a/ember-app/app/mixins/subscriptions/board.js
+++ b/ember-app/app/mixins/subscriptions/board.js
@@ -129,7 +129,7 @@ var BoardSubscriptionMixin = Ember.Mixin.create({
       var copy = `${message.actor.login} changed #${message.issue.number}'s labels`;
       this.get("flashMessages").info(copy);
     }, {time: 5000}),
-    issueUnlabled: sortedQueue(function(message) {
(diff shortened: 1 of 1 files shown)


dahlbyk:
> Due to a type

I hope this typo was on purpose :stuck_out_tongue_winking_eye:

ember-app/app/mixins/subscriptions/board.js
```diff
@@ -129,7 +129,7 @@ var BoardSubscriptionMixin = Ember.Mixin.create({
       var copy = `${message.actor.login} changed #${message.issue.number}'s labels`;
       this.get("flashMessages").info(copy);
     }, {time: 5000}),
-    issueUnlabled: sortedQueue(function(message) {
+    issueUnlabeled: sortedQueue(function(message) {
```
drusellers: :+1:

PR #270, opened 2016-04-14
Adding a check for the 'past_due' status

rauhryan:
Stripe set the subscription status to 'past_due' before it cancels an account. 

Adds a check for that status to allow past_due customers to still access their boards


lib/hu_board/account_helpers.rb
```diff
@@ -25,7 +25,7 @@ def subscription_active?(customer)
       cus = customer && customer[:stripe][:customer]
       if cus && cus[:subscriptions][:total_count] > 0
         sub = cus[:subscriptions][:data][0]
-        return sub[:status] == "active" || sub[:status] == "trialing"
+        return sub[:status] == "active" || sub[:status] == "trialing" || sub[:status] == 'past_due'
```
dahlbyk: Should we use the `%{ trialing past_due }.include?(plan[:status])` syntax here, too?

lib/hu_board/account_helpers.rb
```diff
@@ -25,7 +25,7 @@ def subscription_active?(customer)
       cus = customer && customer[:stripe][:customer]
       if cus && cus[:subscriptions][:total_count] > 0
         sub = cus[:subscriptions][:data][0]
-        return sub[:status] == "active" || sub[:status] == "trialing" || sub[:status] == 'past_due'
+        return %{ active trialing past_due }.include?(plan[:status])
```
dahlbyk: `sub[:status]`?

dahlbyk:
Does https://github.com/huboard/huboard-web/blob/master/vendor/engines/saas/app/helpers/saas/account_helpers.rb need to be updated also/instead?

dahlbyk:
Demo site checks out for me. :shipit:?

PR #271, opened 2016-04-15
Fixes client-side closed issue events causing exceptions after rails upgrade

discorick:
The previous rails upgrade (https://github.com/huboard/huboard-web/pull/237) changed the way ActiveJob serializes hashes, this cause an unexpected breakage in this job causing the actor to be null. 

Closes https://github.com/huboard/huboard/issues/647

### DIFF
--- a/app/controllers/api/issues_controller.rb
+++ b/app/controllers/api/issues_controller.rb
@@ -99,7 +99,7 @@ module Api
         message = {
           :issue => @issue,
           'action_controller.params' => {'correlationId' => params['correlationId']},
-          :current_user => current_user.attribs || {}
+          'current_user' => current_user.attribs || {}
         }
         generate_issue_event(data['state'], message)
       end


app/controllers/api/issues_controller.rb
```diff
@@ -99,7 +99,7 @@ def drag_card
         message = {
           :issue => @issue,
```
dahlbyk: So why does this work if `:current_user` doesn't?

dahlbyk:
Is there any reasonable way to write a test for this?

Should we also guard client-side against a missing `actor`, just in case?

discorick:
https://github.com/huboard/huboard-web/pull/271

discorick:
It's really a one off, pretty much any event missing an actor would cause an exception - we could probably write some testing around the `IssueEventJob` but it'll be non-trivial.

PR #273, opened 2016-04-18
prevents exceptions when Segment is not configured

rauhryan:


vendor/engines/saas/lib/saas/null_analytics.rb
```diff
@@ -0,0 +1,10 @@
+module Saas
```
discorick: I think this should live in: `lib/analytics/adapters` as technically its just a null adapter

discorick:
When do these exceptions occur? SaaS engine shouldn't get loaded for OSS or Enterprise, silent analytics fails in development seems like it could be dangerous.

dahlbyk:
> silent analytics fails in development seems like it could be dangerous.

Could have the null adapter log the method and payload, so it's not silent?

rauhryan:
This happens because there is no default adapter

If you have Saas enabled and Segment disabled, the adapter is nil

rauhryan:
Now that I think about it, it's better to create a Rails logger adapter and set it as the default

vendor/engines/saas/lib/saas/engine.rb
```diff
@@ -15,6 +15,8 @@ class Engine < ::Rails::Engine
         Analytics::Core.adapter = Segment::Analytics.new({
           write_key: ENV["SEGMENTIO_KEY"]
         })
+      else
+        Analytics::Core.adapter = NullAnalytics.new
```
dahlbyk: Rather than require that `adapter` be explicitly set, what if `Analytics::Core` initialized itself with a null or logging adapter by default?

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,3 @@
+Sidekiq.configure_server do |config|
```
discorick: Whats the difference between this guy and the the Rails.logger ?

discorick:
> Now that I think about it, it's better to create a Rails logger adapter and set it as the default

I like!

config/initializers/sidekiq.rb
rauhryan: Nothing they both use STDOUT, it's mainly so that any calls to Rails.logger inside a job put to the same logger. It's so that we can use Rails.logger in the LoggerAdapter

It's just a pointer reference thing basically, Sidekiq respects all of the rails initializers and can execute lib code that might use Rails.logger

In sort, I want to just use Rails.logger and have it show up in Sidekiq

dahlbyk:
Much better. :clap:

PR #275, opened 2016-04-18
Various errors and things that need to be cleaned up in Enterprise

rauhryan:
- [x] disables raygun reporting unless Rails is production and HuBoard is in production 
 - [x] only pulls in stripe when needed
 - [x] makes sure development runs without the Saas engine included

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ gem 'sidekiq', "~> 3.0"
 gem 'saas','0.0.1', require: nil, path: "vendor/engines/saas"
 gem 'faye_extensions','0.0.1', path: 'vendor/gems/faye_extensions'
 gem 'responders'
-gem 'stripe-rails'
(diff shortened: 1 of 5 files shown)


discorick:
Does the fake api key kick up any dirt on exceptions ?

rauhryan:
> Does the fake api key kick up any dirt on exceptions ?

no

discorick:
>  makes sure development runs without the Saas engine included

I think im already doing this but does it mean we need to run development with `HUBOARD_ENV == 'production'`?

rauhryan:
if you want the Saas engine to load, `HUBOARD_ENV` needs to be set to production

dahlbyk:
> if you want the Saas engine to load, `HUBOARD_ENV` needs to be set to production

At some point we should add an internal `saas` synonym for `production` and phase the latter out over time.

PR #276, opened 2016-04-18
Adds a conventional job to pick up the new event from GitHub

rauhryan:


discorick:
Tested, this looks good to me, we can work on the client-side handling later.

PR #90, opened 2015-09-01
Fetches new title info when an issue is opened, adds issue title chan…

discorick:
…ged activity events for reference

PR #278, opened 2016-04-20
Syncs board after a period of inactivity

### LINKED ISSUE #226: Lost Realtime Events
Under certain circumstances real time events can be lost causing boards to fall out if sync with GitHub. 

One obvious reproduction of this is when a browser is closed and reopened later, a saved HuBoard tab will reopen to the last know state, but the events required to sync up are long gone - this can be chaotic for users during morning stand ups and meetings. 

I suspect (but have not confirmed) this can also happen when a tab is either left unattended for long periods of time, or a system has been left in sleep mode. One of our users reports:

> There is something seriously wrong with Huboard's synching with GitHub. Even when I'm actively using Huboard it does not resynch my browser session with GitHub. I'm constantly finding myself in situations where my guys are asking me to look at a bug and when I query for it, it doesn't show up in Huboard. I have to do a full refresh before I will find it. Really frustrating.

Also Relates to missing Events: https://github.com/huboard/huboard/issues/598
Closes https://github.com/huboard/huboard/issues/214

Task List:

- [x] Implement Sticky Toasts, and Stateful Progress Toasts
- [ ] After a period of inactivity, run a diff on the browser on focus to determine if any events have been missed
    - Track the last X number of events processed
    - If events are missing, and not past the X threshold, fetch the missed events and play them on the client
    - Disable board use during any syncing activity
  - If events are missing and are past X threshold, prompt for page refresh

discorick:
Syncs the boards issues after a period of inactivity (currently 1 hour)

- Implements a clientside `BrowserSessionService` which currently tracks browser focus and blurs and events on them using `Ember.Evented`, this is injected into all controllers and components
- Implements `BoardSyncService` which currently can sync the boards issues

TODO:
- [x] Implement an Ember route to force a sync (good for testing)
  - Example: `https://huboard.com/huboard/huboard#/sync-issues`
- [x] Test on Firefox
- [x] Test on IE (on windows 10 huboard is broken :( )
- [x] Test on Chrome

<s> Implement failure handlers (i.e issue count >= 100) </s> I Don't think this is going to be necessary with highly frequent syncs
<s> If the logged in validation fails, throw unauthorized (maybe a seperate PR)</s>

Should Fix #226 & alleviate https://github.com/huboard/huboard/issues/214


config/routes.rb
```diff
@@ -39,6 +39,7 @@
 
 
   namespace :api do
+    get 'logged_in' => 'api#logged_in'
```
drusellers: NIT: `_` in a URL are 👎  prefer `-`

discorick: Is that a common convention (honest question) ? Easy change to make

ember-app/app/services/board-syncing.js
```diff
@@ -0,0 +1,50 @@
+import Ember from 'ember';
+
+var BoardSyncingService = Ember.Service.extend({
+
+  //Sync Notifier
+  flashMessages: Ember.inject.service(),
+  syncFlashNotifier: function(){
+    if(this.get('syncInProgress')){
+      this.get('flashMessages').add(this.messageData);
```
discorick: It occurs to me that `this.messageData` should be cloned lest we modify the original `messageData` object later.

config/routes.rb
drusellers: yeah, it is for me. It goes back a pretty long ways to HTML webpage dev days - imagine http://www.bob.com/this_is_awesome where the default link underlining is in play

image


vs

image

app/controllers/api/api_controller.rb
```diff
@@ -1,5 +1,13 @@
 module Api
   class ApiController < ApplicationController
     skip_before_action :check_account
+
+    def logged_in
+      if logged_in?
+        return render json: {success: true}
+      else
+        return render json: {success: false, status: 403}
```
drusellers: So, this is just a discussion point. you request was technically successful 200 - you are returning the correct payload etc. Also this url is a query, for a pure RESTafarian- 💩 head - should it be `~/session?exist` kind of thing? again, this is purely surface stuff.

config/routes.rb
drusellers: Also this url is a query, for a pure RESTafarian- 💩 head - should it be `~/session?exist` kind of thing? again, this is purely surface stuff.

ember-app/app/controllers/application.js
```diff
@@ -24,6 +24,25 @@ var ApplicationController = Ember.Controller.extend(
...
+  //Browser Session Checker
+  boardSyncing: Ember.inject.service(),
+  checkBrowserSession: function(){
+    var lastFocus = this.get('browser-session.lastFocus');
+    var _self = this;
+    if(lastFocus >= _self.browserCheckInterval){
+      var since = new Date(new Date().getTime() - _self.browserCheckInterval);
+      this.validateCredentials().success((response)=>{
+        _self.get('boardSyncing').syncIssues(_self.get('model.board'), {since: since.toISOString()});
+      }).fail((error)=>{
```
dahlbyk: Rather than separately/explicitly check if the user is logged in here, it seems like we should just handle a 403 when we attempt to fetch latest. More generally, could we use [`jQuery.ajaxSetup`](https://api.jquery.com/jquery.ajaxsetup/) (or a more Ember-y equivalent?) to do something like:

```javascript
$.ajaxSetup({
  statusCode: {
    403: function() {
      alert( "Access denied! Do you need to [login|grant private access]?" );
    },
    500: function() {
      alert( "Oh noes, something went wrong! Try again?" );
    }
  }
});
```

(With better UX than alerts, of course.)

dahlbyk:
> Grey out board during sync ?

Is there some way to make it read-only?

app/controllers/api/api_controller.rb
discorick: Im rethinking this right now... maybe logged_in isn't the best option.

ember-app/app/controllers/application.js
discorick: We already have a global ajax handler picking up failed requests and transition to a "unauthorized" modal.  User login may be the wrong criteria, I honestly didn't think that part of it through too much.

I would like to avoid using ajax catch-alls as much as possible, I see this feature as a way to "latch" on focus and fail before any unauthorized tasks are performed to begin with.

discorick:
>> Grey out board during sync ?

>Is there some way to make it read-only?

We have a CSS overlay (like on issue fullscreen) that I think would be a good touch.

dahlbyk:
> > > Grey out board during sync ?
> 
> > Is there some way to make it read-only?
> 
> We have a CSS overlay (like on issue fullscreen) that I think would be a good touch.

Well I was thinking read-only but otherwise functional. Let me click on issues to see detail (and edit from there, since we always* hit GitHub for latest), but don't let me rearrange stuff in column view.

* right?

discorick:
> * right?

Only in partial, see: https://github.com/huboard/huboard-web/pull/90 (I need to rewrite this to take advantage of #276 which should actually let us start keeping issues completely in-sync)

> Let me click on issues to see detail (and edit ...

I am nervous about pushing the limits of the Ember run loop here. If you happen to make an edit on an issue that is being synced within the bounds of the same run loop, Ember will throw either a deprecation or possibly worse an exception.

ember-app/app/controllers/application.js
dahlbyk: > I see this feature as a way to "latch" on focus and fail before any unauthorized tasks are performed to begin with.

I'm not sure what that actually gains us, though. Server-side, we're checking `logged_in?` in either case, we just have to make two requests if we have an explicit login check.

I'm curious what scenarios we have (or might have) in the app that would require different experiences when a 403 or 500 are encountered?

discorick: > I'm curious what scenarios we have (or might have) in the app that would require different experiences when a 403 or 500 are encountered?

Not entirely clear on the question. A session token can get stale, in which case all actions on the board will cause 403's. We handle this with the unauthorized modal right now, but it would be nice to let users know they need to "relogin" before they attempt to even do something that will fail

dahlbyk:
> If you happen to make an edit on an issue that is being synced within the bounds of the same run loop, Ember will throw either a deprecation or possibly worse an exception.

Hrm... would we be able to compare last-modified timestamps before actually updating an issue from the sync payload? Strictly speaking, the same condition seems possible from an incoming event while the sync is happening.

I'm content with completely disabling the board while syncing, but I do think it would be preferable for the sync to less intrusive. My 2¢.

ember-app/app/controllers/application.js
dahlbyk: Right. My question is: is there any situation in the app where we wouldn't want to handle a 403 with a unified UX to give a chance to restore authorization?

discorick:
After 30 Seconds of no focus:

issue syncing 30 second throttle

dahlbyk:
May just be me, but the sync message is almost too distracting. Once you know there's a sync process, the notification really isn't worth drawing the eye as much as the purple box does. Maybe...

sync-notification

discorick:
I am not sure, thats an interesting thought... Probably outside the scope of this PR though (its flash message styling)?

dahlbyk:
> Probably outside the scope of this PR though (its flash message styling)?

Eh, I would rather see the change here if there is agreement that subtler is better.

drusellers:
> May just be me, but the sync message is almost too distracting. Once you know there's a sync process, the notification really isn't worth drawing the eye as much as the purple box does.

:+1:

discorick:
@dahlbyk @cwramsey @drusellers what do you think?

subtle issue syncing style

cwramsey:
It does look better, but I think just setting the background to `transparent` and nullifying any `border` attributes would get it out of the way even more. You could even set the X button to `opacity: 0;` though maybe not in case the sync somehow takes forever if github is not being very fast.

dahlbyk:
@discorick did raise a good point that simple transparency won't work when the board has been scrolled (at least until we're able to fix the header):
sync-notification2

I just pushed up a commit to try out using a semi-transparent drop shadow that matches the background:
sync-notification3

But not sure if the shadow is really necessary?

I also noticed that the header has a subtle gradient from F3F3F3 to F9F9F9, which makes the non-transparent background not an exact match. I'm wondering if such a subtle gradient is worth keeping altogether?

discorick:
I can live with no shadow, the reason I left it is like you mentioned - the gradient on the navbar makes it look funny without any additional definition.

discorick:
Before we merge this, I have been thinking about thresholds over which a sync would be impractical and a "refresh" recommended toast should be sent. Does 24's of no board focus seem reasonable?

cwramsey:
@discorick imo, 24 hours seems like a lot. I'd set it at like 2 hours maybe. I don't have any proof as to why that's better, but a lot can happen in that timeframe and if it's not automatically caught up (like if the connection is lost after sleeping your computer or something) a sync would be good then.

discorick:
@cwramsey the 24 hour interval is for forced refresh, thats the point where we say "sorry you need to hit the refresh button now"

ember-app/app/services/board-syncing.js
```diff
@@ -0,0 +1,75 @@
...
+  //Sync Notifier
+  flashMessages: Ember.inject.service(),
+  syncFlashNotifier: function(){
+    if(this.get('syncInProgress')){
+      this.get('flashMessages').add(this.messageData());
+    } else {
+      var flash = this.get('flashMessages.queue').find((flash)=>{
+        return flash.identifier === 'sync-message';
+      });
+      Ember.set(flash.progress, 'status', false);
```
dahlbyk: Is there any need to guard against `flash` being `null` here?

PR #280, opened 2016-04-25
[WIP] Make the columns scrollable with horizontal scroll on the milestones view

rauhryan:

PR #284, opened 2016-05-03
[WIP] Contrains adding new links to a single run loop

discorick:
Testing to see if render times for linking large repos can be reduced by constraining to a single run loop

### DIFF
--- a/ember-app/app/controllers/settings/links/index.js
+++ b/ember-app/app/controllers/settings/links/index.js
@@ -42,9 +42,11 @@ var SettingsLinksIndexController = Ember.Controller.extend({
       var model = this.get("model.repo");
       this.set("isDisabled", true);
 
-      model.createLink(name, _self.get("issue_filter")).then(()=> {
-        _self.resetForm();
-      }, (jqXHR)=> _self.errorHandler(jqXHR) )
+      Ember.run.once(()=>{
+        model.createLink(name, _self.get("issue_filter")).then(()=> {
+          _self.resetForm();
+        }, (jqXHR)=> _self.errorHandler(jqXHR) );
+      });
     },
     validate: function(){
       var _self = this;
--- a/ember-app/app/controllers/settings/links/index.js
+++ b/ember-app/app/controllers/settings/links/index.js
@@ -42,11 +42,9 @@ var SettingsLinksIndexController = Ember.Controller.extend({
       var model = this.get("model.repo");
       this.set("isDisabled", true);
 
-      Ember.run.once(()=>{
-        model.createLink(name, _self.get("issue_filter")).then(()=> {
-          _self.resetForm();
-        }, (jqXHR)=> _self.errorHandler(jqXHR) );
-      });
+      model.createLink(name, _self.get("issue_filter")).then(()=> {
+        _self.resetForm();
+      }, (jqXHR)=> _self.errorHandler(jqXHR) )
     },
     validate: function(){
       var _self = this;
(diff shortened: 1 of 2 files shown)


discorick:
Ok, for large repos (I tested on dotnet/roslyn) this change can double the render speed! In this particular case render went from 40 Seconds on link down to just over 20 seconds on link. 

Tested on:
 https://huboard-rails-pr-284.herokuapp.com/huboard-testing/test3#/settings/links

dahlbyk:
Fun fact: unless you hit the back button, the only way out of the board is to click on repo in the `org > repo` breadcrumb at the top. Which re-renders the board anyway. What am I missing?

This seems to actually make the re-render worse (on my machine) when adding a link to aspnet/mvc:
chrome_2016-05-05_01-34-48

dahlbyk:
> Fun fact: unless you hit the back button, the only way out of the board is to click on repo in the `org > repo` breadcrumb at the top. Which re-renders the board anyway. What am I missing?

To be more specific, clicking on the repo in the breadcrumb causes a _full_ reload. Clicking on Milestones or Tasks will re-render.

It seems to me that we should continue to fetch the linked repo's details when the link is created, but prevent Ember from trying to render it at all (since clicking anywhere just re-renders anyway).

Other nitpicks on the settings experience, while I'm here:
1. There should be some sort of a Close button, maybe an X where the Settings icon is?
2. We should hijack the repo breadcrumb link to function as a link to the Tasks board.

discorick:
> It seems to me that we should continue to fetch the linked repo's details when the link is created, but prevent Ember from trying to render it at all (since clicking anywhere just re-renders anyway)

Given the complexities of holding those objects in a Proxied location in memory, before "letting them loose" into the computed arrays, just prompting for a refresh before going back to the task view would be simpler.

I documented your nitpicks here: https://github.com/huboard/huboard/issues/657

dahlbyk:
Another change in behavior, I believe because we're initially saying the board is loaded with `columns` set to `[]`:
link

Production:
link-prod

discorick:
Yeah - this spike was only concerned about perf, I haven't even looked at the UX implications yet.

dahlbyk:
> This seems to actually make the re-render worse (on my machine) when adding a link to aspnet/mvc

Any thoughts on this? You had mentioned it may be related to label counts on issues?

discorick:
> Any thoughts on this? You had mentioned it may be related to label counts on issues?

Label counts on repos, our filtering becomes inefficient on repos with many many labels, still need to test the theory out

discorick:
I think this confirms that high volumes of labels are a huge factor in these slow downs, using rauhryan/skipping_stones_repo (many labels) the render performance is awful, other repos with comparable issues counts but less labels render up to 1000% faster:

image

dahlbyk:
> I think this confirms that high volumes of labels are a huge factor in these slow downs, using rauhryan/skipping_stones_repo (many labels) the render performance is awful, other repos with comparable issues counts but less labels render up to 1000% faster:

I'm curious if it's due to issue-label associations or just raw label count overall. Create a test repo with lots of labels and few (unlabeled) issues?

discorick:
> I'm curious if it's due to issue-label associations or just raw label count overall. Create a test repo with lots of labels and few (unlabeled) issues?

I actually tested removing issues from the picture altogether (not bringing them into the app at all), it's purely a labels problem. Still need to track down why setting so many labels on a repo causes the slowdown.

discorick:
Closing in favor of https://github.com/huboard/huboard-web/pull/280, it turns out the CSS View was the root of some evils

Issue #229, opened 2016-03-10
Move dashboard to /dashboard


dahlbyk:
I like Segment's approach - landing on the home page, even if logged in, shows their marketing site. Clicking Login immediately takes you to your workspaces.

I propose that we:

* Add `/dashboard` route for the dashboard.
* Remove the logic to show the dashboard on `/` if you're authenticated.
* Make `/login` redirect to `/dashboard` if you're already authenticated.

drusellers:
Github.com does something similar too don't they?

discorick:
Moving to 'next release' as this is almost complete already

PR #285, opened 2016-05-03
Improved Login UX

discorick:
Improved UX for logging in:

- Clearer copy on the home/login/pricing pages 
- Restyles the login button (octocat!)
- Defaults to public oauth scope on initial login, offers private upgrade later on

Merges the previously spiked out UX improvements from #177 & #181


dahlbyk:
@petey-rock actually has a test queued up to drop the Private side of the Login page. Drop that change from this PR until we can confirm it helps overall login click-through?

I do think we should ship the public-only auth experience before we enable that test, though.

discorick:
This PR removes any UX leading to that page altogether and moves users directly to the dashboard... Do you think it's worth the time doing anymore testing or work on deprecated view?

dahlbyk:
Well I'd like to prove the theory that having public/private options lead to confusion and more abandons, but I suppose can get equivalent confirmation from Google Analytics. Good call, carry on.

app/views/login/index.html.erb
```diff
@@ -41,131 +41,62 @@
...
+        <dt>Commit status</dt>
+        <dd>
+          <img src="<%= image_path 'marketing/permissions-read.png' %>" alt="Read" title="Read">
+          <img src="<%= image_path 'marketing/permissions-write.png' %>" alt="Write" title="Write">
+        </dd>
+        <dt>Notifications</dt>
+      </dl>
+      <a href="/login/public" class="btn btn--flat">
+        <img class="btn-octocat" src="<%= image_path 'GitHub-Mark-Light-32px.png' %>" width="16px"> Sign in with GitHub
+      </a>
```
dahlbyk: Button should move to the top.

dahlbyk:
> > This PR removes any UX leading to that page altogether and moves users directly to the dashboard... Do you think it's worth the time doing anymore testing or work on deprecated view?
> 
> Well I'd like to prove the theory that having public/private options lead to confusion and more abandons, but I suppose can get equivalent confirmation from Google Analytics. Good call, carry on.

Sorry, I had not fully reacquainted myself with the changes here (can update the issue description to enumerate them?).

The assumption here is that the `/login` page doesn't have any value, so we're better off sending directly to GitHub. I'm not entirely convinced that's the case, but would love to be proved wrong. To get the best understanding of our auth flow, I propose that we:

1. Implement #229 to make our lives easier in identifying home vs dashboard hits in analytics/testing.
2. Drop the `/login` :point_right: `/login/public` direct link changes here.
3. Proceed with making `/login` simpler:
   * Drop Private column
   * Drop redundant copy
    * Add "Why do you need this?" copy linking to [newly updated](https://github.com/huboard/huboard/wiki/General-Support/_compare/a646f24e0eb2e5745c6778f67574f15ac92d0eca...489104bef69ff588e87a12197431a7f8beb7f016) [FAQ](https://github.com/huboard/huboard/wiki/General-Support#read-write)
4. Use Optimizely to test if linking straight to `/login/public` results in as many auths as the improved `/login`.

app/views/login/index.html.erb
dahlbyk: Also, this whole block should be correctly indented.

app/views/marketing/marketing.html.erb
```diff
@@ -8,7 +8,7 @@
 		<p class="hero__copy">
 			HuBoard is built from the ground up using the GitHub public API. HuBoard issues are GitHub issues, you will never have to deal with synchronization problems. Keep issues where they belong, in the repository with your code!
 		</p>
-		<a href="/login" class="btn hero__btn">Sign up for a 14 day free trial <i class="icon icon--lt"></i></a>
+		<a href="/login/public" class="btn hero__btn">Try HuBoard for free! <i class="icon icon--lt"></i></a>
```
dahlbyk: Even if we end up keeping the header change to `/login/public`, it seems wise to me that "Try Us" type links would still go through `/login` to give us a chance to clarify what the following GitHub auth screen means.

discorick:
> Implement #229 to make our lives easier in identifying home vs dashboard hits in analytics/testing.

I assume you intend this as a follow up task?

> 2. Drop the `/login` :point_right: `/login/public` direct link changes here.
3. Proceed with making `/login` simpler:
   * Drop Private column
   * Drop redundant copy
    * Add "Why do you need this?" copy linking to [newly updated](https://github.com/huboard/huboard/wiki/General-Support/_compare/a646f24e0eb2e5745c6778f67574f15ac92d0eca...489104bef69ff588e87a12197431a7f8beb7f016) [FAQ](https://github.com/huboard/huboard/wiki/General-Support#read-write)
4. Use Optimizely to test if linking straight to `/login/public` results in as many auths as the improved `/login`.

Okay, I am on board with this!

discorick:
`/` Before: 
image

`/` After:
image

`/login` Before:
image

`/login` After:
image

`/pricing` Before:
image

`/pricing After:`
image

dahlbyk:
Looking good! Random thoughts:

* I tried to click on the purple text, since that's our link color. Maybe black?
  firefox_2016-05-05_02-00-54

* Not sure if I prefer overlap or touching.
   firefox_2016-05-05_02-05-54

* There's some weird stuff going on with the responsive breakpoints on `/login`:
  breakpoints

dahlbyk:
> > Implement #229 to make our lives easier in identifying home vs dashboard hits in analytics/testing.
> 
> I assume you intend this as a follow up task?

Sort of. But I got curious, so.... https://github.com/huboard/huboard-web/pull/287

app/assets/stylesheets/_repos.scss
```diff
@@ -66,4 +67,55 @@ ul.repos li a.btn {
...
+  &.public-only-hero {
+    position: relative;
+    bottom: 10px;
+    text-align: center;
+  }
 
+  &.public-only-hero:before {
+    content: '';
+    width: 0px; height: 5px;
+    margin: -20px 25px 0px auto;
```
dahlbyk: I think there's a 1px gap between the arrow and the border. Try `-19px`for top?

discorick:
> "Try HuBoard for free!" vs "Try HuBoard For Free!"? We're not

Are you saying you don't think we allow people to try HuBoard for free? (I disagree)

> Not sure if I prefer overlap or touching.

I like the overlap, it leaves less of a gap between the warning banner and the container.

dahlbyk:
> > "Try HuBoard for free!" vs "Try HuBoard For Free!"? We're not
> 
> Are you saying you don't think we allow people to try HuBoard for free? (I disagree)

I never finished that thought, sorry. I was going to say "We're not consistent in our use of casing," but on further inspection I think we generally are consistent in not using Title Case Everywhere. So :+1: as-is.

dahlbyk:
CSS adjustments so far look good. Do we want to adjust the menu breakpoint to avoid this wrapping?
image

Based on https://github.com/huboard/huboard-web/issues/172#issuecomment-217196057, I think we need to revert the changes that will force users through `/login/public` for now.

discorick:
Hmm theres a weird gap between when the responsive layout kicks in and the logo pops up like that. I must have thrown of a calculation somewhere I'll look into it, good catch.

dahlbyk:
> I must have thrown of a calculation somewhere I'll look into it, good catch.

I believe it's simply because the Sign In button got wider.

discorick:
Hey @rauhryan , take a look at 0061d6c, I would expect that the callback here (which is working) to have the `current_user` context but it doesn't. Is there something I am missing here ?

app/assets/stylesheets/marketing/modules/_variables.scss
```diff
@@ -15,7 +15,7 @@ $max-width: 				1001px;
  * Breakpoints
  */
 $short: 					new-breakpoint(max-width 22.5em 4);     // 360px
-$tall: 						new-breakpoint(max-width 47.8em 12);      // 768px
+$tall: 						new-breakpoint(max-width 52.94em 12);      // 847px
```
dahlbyk: I'm concerned that this will affect more than just the header wrapping. Maybe better to adjust nav one-off?

dahlbyk:
So we're going to fix All The Things™ here, rather than shipping the incremental improvements first?

app/assets/stylesheets/marketing/modules/_variables.scss
discorick: I would be surprised if thats the case, this calculation is the _exact_ width required to keep the logo from "bumping up"

dahlbyk:
CSS seems to be flowing well from all my testing. Anything else here?

discorick:
I think thats a wrap!

PR #287, opened 2016-05-05
Add /dashboard route

dahlbyk:
In support of https://github.com/huboard/huboard-web/pull/285, here's most of #229:

☑ Add /dashboard route for the dashboard.
☒ Remove the logic to show the dashboard on / if you're authenticated.
☑ Make /login redirect to /dashboard if you're already authenticated.

We can kick the `/` change down the road... I'm not 100% sold on that.

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -8,12 +8,12 @@ class LoginController < ApplicationController
     github_authenticate! :default
     @user = gh.user
     @emails = @user.emails.all
-    redirect_to params[:redirect_to] || "/"
+    redirect_to params[:redirect_to] || "/dashboard"
   end
   def private
     github_authenticate! :private
     @user = gh.user
     @emails = @user.emails.all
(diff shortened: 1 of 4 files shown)


discorick:
Ok this looks easier than I thought. Could you make sure the hamburger links to `/dashboard` ? Given that change I can't think of any reason not to go ahead and:

> Remove the logic to show the dashboard on / if you're authenticated.

>We can kick the / change down the road... I'm not 100% sold on that

Are you saying your not sold on the change? Or you are not sold on kicking the can on it ?

dahlbyk:
> Could you make sure the hamburger links to `/dashboard` ?

Done.

> Are you saying your not sold on the change? Or you are not sold on kicking the can on it ?

Not sold. Quite a few UX implications we would need to work through, and a relatively low value change at the moment (IMO).

discorick:
Ok, that sounds good to me. :shipit: as is

PR #292, opened 2016-05-10
Guard against nil scope

dahlbyk:
More durable guard than ac3fe55dee766016c2bd240b4de15cf9be1fdbeb, I think?


dahlbyk:
Superseded by #293

PR #294, opened 2016-05-11
adds a puma worker killer to test out memory problems on heroku

rauhryan:


dahlbyk:
Can you give some background on what you're seeing to motivate this? Is this just a test, or...?

Related links from reading up on puma worker killer:

* https://devcenter.heroku.com/articles/ruby-memory-use
* https://www.nateberkopec.com/2015/07/22/secrets-to-speedy-ruby-apps-on-heroku.html

PR #296, opened 2016-05-11
Catches unhandled board sync 404's

### LINKED ISSUE #295: Message: no implicit conversion of String into Integer
We have recently had several instances of:

> Message: no implicit conversion of String into Integer
>
> /app/lib/bridge/github/issues.rb:267 []=
> /app/lib/bridge/github/issues.rb:267 extended
> /app/lib/bridge/github/issues.rb:18 extend
> /app/lib/bridge/github/issues.rb:18 block in issues
> /app/vendor/bundle/ruby/2.3.0/gems/ghee-0.14.19/lib/ghee/resource_proxy.rb:51 each
> /app/vendor/bundle/ruby/2.3.0/gems/ghee-0.14.19/lib/ghee/resource_proxy.rb:51 method_missing
> /app/lib/bridge/github/issues.rb:17 issues
> /app/app/controllers/api/issues_controller.rb:11 issues

It's not immediately obvious to me what the problem might be:
https://github.com/huboard/huboard-web/blob/657abcd8e17b070775c4b3a4c8f1b05a35d1621b/lib/bridge/github/issues.rb#L266-L271

https://app.raygun.io/crashreporting/7lc2yw/errors/1334243193
https://app.raygun.io/crashreporting/7lc2yw/errors/1325060943
https://app.raygun.io/crashreporting/7lc2yw/errors/1309424285

discorick:
When a board attempts to sync and there is no session, Ghee returns a 404 Not Found. Adds a handler to properly bubble this up to the client. 

Currently the UX for this looks like:
image

This fixes the `Message: no implicit conversion of String into Integer` popping into Raygun all day.
Closes #295


dahlbyk:
Is there a way we could try `/login/github` to nudge the session before we ask for reauth?

discorick:
> Is there a way we could try `/login/github` to nudge the session before we ask for reauth?

Could get a little messy, as the board sync has already failed - yes we could try to refresh the session automatically, but the UI would still be notifying the user that the board sync failed anyway.

rauhryan:
If we are going to edit this part of the code base, I'd like to take care a few UX warts 

  - [ ] the "modal error screen" should **NOT** be routable anymore
  - [ ] change the logic of the allow access button to use the new login flow
  - [ ] clicking the allow access button, authenticates and properly returns you back to the board and **NOT** the homepage

dahlbyk:
I'm more than content to file this under our "Improve Login Flow" initiative.

discorick:
> clicking the allow access button, authenticates and properly returns you back to the board and NOT the homepage

I think I fixed that bug some time back, it currently returns back to the board on production

Good call on the other two items!

discorick:
I made some tweaks to the unauthorized UX, the flash message rather than prompting for a refresh keeps seraching suggesting (hopefully) its in a suspended state: 

image

rauhryan:
I tested the functionality and the Re-enable button logged me back in and redirected 👍 

Quick win would be to add the same redirect logic to the Sign in button

rauhryan_skipping_stones_huboard_and_huboard_-_github_issues_made_awesome__and_roxanne_wallin

After that, I think we :shipit:

dahlbyk:
> Quick win would be to add the same redirect logic to the Sign in button

:+1:

Love this change, nice work!

PR #297, opened 2016-05-11
Fixes double assignee events from blocking assignments

discorick:
Conversation & Fixes https://github.com/huboard/huboard/issues/658


rauhryan:
looks solid :shipit:

PR #255, opened 2016-03-29
Implements pull_request web hook to health checks and client side events

rauhryan:
converts PR into an issue payload and sends up the
event as an issue event. 

Seems to work 

Closes #231
Closes huboard/huboard#608
Related to huboard/huboard#598

PR #299, opened 2016-05-13
Differentiate new PRs from issues in notifications

dahlbyk:
Follow-up from #255

Fixes this:
slack_2016-05-13_15-03-16

### DIFF
--- a/app/jobs/api/issues_open_issue_job.rb
+++ b/app/jobs/api/issues_open_issue_job.rb
@@ -6,6 +6,7 @@ module Api
 
     def payload(params)
       {
+        pull_request: params[:pull_request],
         issue: params[:issue]
       }
     end
--- a/app/controllers/api/webhooks_controller.rb
+++ b/app/controllers/api/webhooks_controller.rb
@@ -31,6 +31,7 @@ module Api
       payload[:pull_request].extend(Huboard::Issues::Card).merge!(repo)
 
       message = HashWithIndifferentAccess.new(
+        :pull_request => true,
         :issue => payload[:pull_request],
         :label => payload[:label],
         "action_controller.params" => {},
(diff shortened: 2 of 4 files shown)


dahlbyk:
This isn't working: `mash.payload.pull_request.nil?`. I have once again conflated event and web hook payloads.

Rather than infer PR-ness from `payload.issue` properties (`head`, `base`, `commits_url`), it seems reasonable to explicitly mark PR payloads as such: https://github.com/huboard/huboard-web/pull/299/commits/77b26c34087a30fd54ca38ce46746720efc33ff7.

dahlbyk:
Third time's a charm:
slack_2016-05-13_16-27-40

Ready for review.

app/jobs/api/issues_open_issue_job.rb
```diff
@@ -6,6 +6,7 @@ class IssuesOpenIssueJob < IssueEventJob
 
     def payload(params)
       {
+        pull_request: params[:pull_request],
```
discorick: this could also be written `params[:issue].key?(:pull_request)` if the webhook payload is similar enough to the original issue payload that is.

dahlbyk: I'm adding `:pull_request` to the base payload, not attaching it to the `:issue` hash.

discorick: Oh, I assume in discovery you found that the webhook doesn't carry a PR object ?

dahlbyk: Correct, in `WebhooksController#publish_pull_request_event` we essentially "rename" the `pull_request` part of the payload as `issue`. So I added a simple flag in 77b26c34087a30fd54ca38ce46746720efc33ff7 to identify PRs again:

```diff
        message = HashWithIndifferentAccess.new(
+        :pull_request => true,
         :issue => payload[:pull_request],
         ...
```

This change just passed that value on through.

discorick:
Looks good to me :shipit: ing it

PR #301, opened 2016-05-16
Card Redesign API Prep

discorick:
Taking care of some smaller items to enable upcoming work on Card Redesign/Relationships

Addresses items here: https://github.com/huboard/huboard/issues/661


dahlbyk:
Can we say this closes huboard/HuBoard#661? I don't think we want to block progress on the redesign to implement task progress, we'll just leave a spot for it in the cards.

discorick:
> More importantly, I'm seeing a bunch of duplicate issues in the board details payload!

Ah good catch, I called `all` in the wrong place - I can fix! This is also the same reason the details are missing the reaction payload >.<

lib/bridge/github/issues.rb
```diff
@@ -16,9 +16,9 @@ def issues(label = nil, opts={})
 
       issues_response = gh.issues(params) do |request|
         request.headers["Accept"] = "application/vnd.github.squirrel-girl-preview"
-      end
+      end.all
 
-      issues_response.all.each{
+      issues_response.each{
```
dahlbyk: :ok: I will bite. What is the difference here? I would expect these to be equivalent.

discorick: `gh.issues(params)` returns a `Ghee::Proxy`, `all` initializes the call, for some reason assigning the proxy to `issue_response` and then calling `all` on a subsequent invocation seems to make the proxy double up on the request.

dahlbyk: Is thay a quirk of Ghee's proxy implementation, or something inherent to Ruby itself? Assigning an intermediate variable should never implicitly change behavior. :frowning:

rauhryan: I'm not sure, Ghee does some weird magically stuff sometimes. 

Might be a quirk in Ghee.

I think it's because you have a pointer reference to the proxy object in the old diff, the proxy object must not modify itself or consider itself "fulfilled" so subsequent calls to all would execute again? 

Not sure why that would cause double entries in the array tho

dahlbyk:
> Fixies misplaces all method on issues api call https://github.com/huboard/huboard-web/pull/301/commits/a3c9e1273d1fb40814d2211bff9571c87a93b9fd

Still seeing double, without reactions...?

discorick:
>> Fixies misplaces all method on issues api call https://github.com/huboard/huboard-web/pull/301/commits/a3c9e1273d1fb40814d2211bff9571c87a93b9fd

>Still seeing double, without reactions...?

Weird my machine must have been a weird state, ok i have confirmed this is now fixed, I don't know why but the `all` was the cause. Must be a quirk with Ghee.

dahlbyk:
Now we're missing a bunch of issues. I assume `to_a` is skipping pagination?

discorick:
Which repo are you missing issues on? My test repo was fine, need to be able to repro to make sure .

dahlbyk:
https://huboard-rails-pr-301.herokuapp.com/huboard/huboard

discorick:
For realz @dahlbyk , I think we got I this time :P

dahlbyk:
> For realz @dahlbyk , I think I got it this time :P

Confirmed

gitleca:
Are all these people intelectuals technologicaly speaking??

PR #303, opened 2016-05-18
Add analytics job for /login/github

dahlbyk:
https://github.com/huboard/huboard-web/pull/291 + https://github.com/huboard/huboard-web/pull/232

### DIFF
--- a/vendor/engines/saas/app/jobs/login_github_job.rb
+++ b/vendor/engines/saas/app/jobs/login_github_job.rb
@@ -1,3 +1,0 @@
-class LoginGithubJob < BaseLoginJob
-  action 'github_login'
-end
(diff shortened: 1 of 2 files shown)


dahlbyk:
`/login/github` confirmed in [Segment](https://segment.com/huboard/sources/huboard-development/debugger). `/login/github/authorized` isn't firing since we apparently changed the callback to `/oauth/github/callback`.

That said, I'm not sure we even need to track that explicitly, since you'll either land on `/welcome` or `/dashboard`. Thoughts?

discorick:
Sorry about the delay on this one... yeah we can now assume dashboard is authorized, and welcome another step in the funnel.

PR #306, opened 2016-05-25
Accommodates custom media types

discorick:
This fixes an issue where `body_html` keys are not making their way to the board. 

Fixes regression from https://github.com/huboard/huboard-web/pull/301

### DIFF
--- a/lib/bridge/github/issues.rb
+++ b/lib/bridge/github/issues.rb
@@ -15,7 +15,7 @@ class Huboard
       params = params.merge(labels: label) if label
 
       issues_response = gh.issues(params).all do |request|
-        request.headers["Accept"] = "application/vnd.github.squirrel-girl-preview"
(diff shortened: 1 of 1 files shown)


dahlbyk:
I'm concerned how easy of a regression this would be to repeat. [`full+json` is hard-code in Ghee](https://github.com/huboard/ghee/blob/85525605f26229485969f6d077b491e4345e2541/lib/ghee/connection.rb#L38) and is otherwise invisible. We could test for it being set on `request` here, but there is no guarantee we won't forget next time.

We can probably merge as-is to avoid the regression on our next deploy, but I want to follow up with a solution that yields more confidence. Thoughts?

dahlbyk:
More generally, [GitHub API Media Types](https://developer.github.com/v3/media/) docs suggest the `full+json` suffix doesn't apply globally. We should definitely rethink how Ghee handles `Accept` headers.

PR #309, opened 2016-05-27
Implements references in the issue model

discorick:
Adds:

- `accept(vistor)` to the base client side model
- Uses the `IssueReferencesVisitor` to discover issues based on references found in the body (asynchronously)

Discussion point, how much should we hydrate issues that are closed and not on the board? Currently I am returning the reference, with a state i.e

```javascript
{
  url: 'https://huboard.com/huboard/huboard/issues/1'
  id: 12345,
  text: '#1'.
  state: 'closed'
}
```

Discovered issue references that already exist in memory are returned by their in memory counterparts (a full issue model). 

related to huboard/huboard#666


ember-app/app/visitors/issue/references.js
```diff
@@ -0,0 +1,54 @@
...
+      reject([]);
+    });
+  },
+
+  discoverIssues: function(issue){
+    var references = issue.get('cardRelationships')['issue-references'] || [];
+    var issuesById = issue.get('repo.board.issuesById');
+    return references.map((reference)=>{
+      if(issuesById.hasOwnProperty(reference.id)){
+        return issuesById[reference.id][0];
```
dahlbyk: Was thinking the need for `[0]` was less than ideal - found [`indexBy()`](http://underscorejs.org/#indexBy) for just this case:

>  Just like `groupBy`, but for when you know your keys are unique.

ember-app/app/models/new/board.js
```diff
@@ -92,6 +92,16 @@ var Board = Model.extend({
       return combined;
     }
   }),
+  issuesById: function(){
+    return _.groupBy(this.get('issues'), (issue)=> {
+      return issue.get('id');
+    });
+  }.property('issues.[]'),
```
dahlbyk: To clarify my Ember understanding, this means the the `issuesById` function will be re-run any time the `issues` collection changes (i.e. a message comes in) but otherwise the value will be reused?

discorick: You bet, basically everytime you ask for the value of `issuesById` Ember will check the `issues.[]` array and ensure it has not changed otherwise it recomputes.

ember-app/app/visitors/issue/references.js
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }).compact();
+  },
+  discoverClosedIssues: function(issue){
+    var references = issue.get('cardRelationships')['issue-references'] || [];
+    var discovered_issues = this.discoverIssues(issue);
```
dahlbyk: It seems weird that we're calling `discoverIssues(issue)` for a second time here. It seems to me that the open/closed/missing logic could all happen on a single iteration of `references`, i.e.

```javascript
discoverIssues: function(issue) {
    var references = issue.get('cardRelationships')['issue-references'] || [];
    return references.map(ref => discoverOpenIssue(ref) || discoverClosedIssue(ref) || discoverMissingIssue(ref));
}
```

ember-app/app/models/new/board.js
```diff
@@ -92,6 +92,16 @@ var Board = Model.extend({
...
+  issuesByRepo: function(){
+    return _.groupBy(this.get('issues'), (issue)=> {
```
dahlbyk: It seems like we don't really need the actual issues per repo, just the unique list of `repo.full_name`? If I'm reading this correctly, we can use `countBy` to save on allocation of new arrays to hold the repos' issues.

ember-app/app/visitors/issue/references.js
discorick: I guess it depends how we ultimately want to handle missing issues, closed or otherwise (Create 'best guess' data or actually fetch the issue?) 

`return references.map(ref => discoverOpenIssue(ref) || discoverClosedIssue(ref) || discoverMissingIssue(ref));` 

Would work synchronously, but would potentially invoke 3 promises per reference.

dahlbyk: `discoverMissingIssue` seems the only one that might need a promise. `discoverOpenIssue` and `discoverClosedIssue` should both be quick index lookups, essentially.

discorick: True, I have them all wrapped async at the moment so they don't hold up board/card rendering in any way (avoiding the issues prevalent on the milestones columns)

ember-app/app/models/new/board.js
discorick: So, this is kind of a cheat, but I would like the `issuesByRepo` for other parts of the app as well where I think we could get a performance benefit.

ember-app/app/visitors/issue/references.js
dahlbyk: :+1: for async. How about something like:

```javascript
  visit: function(issue){
    var references = issue.get('cardRelationships')['issue-references'] || [];
    var promises = references.map(this.discoverIssues);

    Ember.RSVP.all(promises).then((refs)=> {
      issue.set('issueReferences', refs);
    });
  },
```

`discoverIssues` would return a promise that either returns the open/closed issue from the current board, or fetches a new one.

dahlbyk:
This is looking pretty good, but I don't think references from a linked board back to the parent board are working (e.g. https://huboard-rails-pr-309.herokuapp.com/huboard/huboard#/issues/104311313).

discorick:
@dahlbyk How did you determine that? (it wont be in the payload of the issue details)

dahlbyk:
> How did you determine that? (it wont be in the payload of the issue details)

I set a breakpoint somewhere in the visitor, probably in `discoverClosedIssue`.

discorick:
> This is looking pretty good, but I don't think references from a linked board back to the parent board are working (e.g. https://huboard-rails-pr-309.herokuapp.com/huboard/huboard#/issues/104311313).

@dahlbyk got it here:  aa0363d

PR #311, opened 2016-06-02
Chains the card extension to fix failed payload merges

discorick:
I have no idea why this works, but the `Ghee::ResourceProxy` will not properly extend the object unless the calls are chained together. This breakage started right here: https://github.com/huboard/huboard-web/pull/301/commits/e359da75c3e13cf166d09a3af4b66b1cf551fa06#diff-55d956adac5de9504d394fc2fa4b22dcR75


dahlbyk:
:shipit:

PR #314, opened 2016-06-03
Fix rare MAX/MIN overflows in issue ordering

discorick:
As reported by a couple users, on rare occasion newly created cards can be bumped part-ways down the column, and cards moves can occasionally cause strange behavior (it will jump to somewhere else on drag)

It turns out the cause of these troubles relate to javascripts `Number.MIN_VALUE/MAX_VALUE` and how HuBoard has calculated moves to persist orders over the years. When an order reaches to point where the MIN/MAX is overwhelmed, the card is no longer moveable, and moves around that card became impossible for JS to compute. 

This PR addresses this in a couple of ways:

- Any current cards on a board that are over/under  `minThreshold: 1e-319` || `maxThreshold: 1e307`  will be algorithmically scaled down to more digestable orders (unfortunately this _will_ shuffle currently affected cards)

- The increments for moving cards up and down a column have been drastically reduced, which should prevent this from ever happening again in the future. 

Shoutout to @dahlbyk who helped me greatly in getting the Maths right on this

### DIFF
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -23,6 +23,7 @@ var CardMoveMixin = Ember.Mixin.create({
     moveToTop: function(issue_below){
       var below_order = issue_below._data[this.get("orderKey")]
       var order = below_order / this.orderMultiplier;
+      if(order <= 0){ order = issue_below.get('data.id') * this.minThreshold; }
       while(order < this.minThreshold){ order *= 10; }
 
       return order;
@@ -30,6 +31,8 @@ var CardMoveMixin = Ember.Mixin.create({
     moveToBottom: function(issue_above){
       var above_order = issue_above._data[this.get("orderKey")];
       var order = above_order * this.orderMultiplier;
+
+      if(order === Infinity){ order = issue_above.get('data.id') / this.maxThreshold; }
       while(order > this.maxThreshold){ order /= 10; }
 
       return order;
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -23,6 +23,7 @@ var CardMoveMixin = Ember.Mixin.create({
     moveToTop: function(issue_below){
       var below_order = issue_below._data[this.get("orderKey")];
       var order = below_order / this.orderMultiplier;
+
       if(order <= 0){ order = issue_below.get('data.id') * this.minThreshold; }
       while(order < this.minThreshold){ order *= 10; }
 
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -21,7 +21,7 @@ var CardMoveMixin = Ember.Mixin.create({
       return (above_order + below_order) / 2;
     },
     moveToTop: function(issue_below){
-      var below_order = issue_below._data[this.get("orderKey")]
+      var below_order = issue_below._data[this.get("orderKey")];
       var order = below_order / this.orderMultiplier;
       if(order <= 0){ order = issue_below.get('data.id') * this.minThreshold; }
       while(order < this.minThreshold){ order *= 10; }
@@ -32,7 +32,7 @@ var CardMoveMixin = Ember.Mixin.create({
       var above_order = issue_above._data[this.get("orderKey")];
       var order = above_order * this.orderMultiplier;
 
-      if(order === Infinity){ order = issue_above.get('data.id') / this.maxThreshold; }
+      if(order === Infinity){ order = this.maxThreshold / 1e10 * issue_above.get('data.id'); }
       while(order > this.maxThreshold){ order /= 10; }
 
       return order;
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -21,10 +21,15 @@ var CardMoveMixin = Ember.Mixin.create({
       return (above_order + below_order) / 2;
     },
     moveToTop: function(issue, issue_below){
-      return (issue_below._data[this.get("orderKey")]) / 2;
+      var order = (issue_below._data[this.get("orderKey")]) / this.orderMultiplier;
+      if(order < (this.orderMultiplier * Number.MIN_VALUE)){
+        return issue_below.get('data.id') * Number.MIN_VALUE;
+      }
+
+      return order;
     },
     moveToBottom: function(issue, issue_above){
-      return issue_above._data[this.get("orderKey")] + 1;
+      return issue_above._data[this.get("orderKey")] * this.orderMultiplier;
     },
     findCard: function(element, column){
       return column.get("cards").find(function(card){
@@ -59,7 +64,8 @@ var CardMoveMixin = Ember.Mixin.create({
       //Adjust based on issue dragging up or down
       if(column_changed){ return 0; }
       return index >= this.data.originIndex ? 1 : 0;
-    }
+    },
+    orderMultiplier: 1.0001
   })
 });
 
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -11,8 +11,8 @@ var CardMoveMixin = Ember.Mixin.create({
     calculateIssueOrder: function(issue_above, issue_below){
       var issue = this.data.card.get("issue.data");
       if(!issue_above && !issue_below){return issue._data.order; }
-      if(!issue_above){ return this.moveToTop(issue, issue_below); }
-      if(!issue_below){ return this.moveToBottom(issue, issue_above); }
+      if(!issue_above){ return this.moveToTop(issue_below); }
+      if(!issue_below){ return this.moveToBottom(issue_above); }
       return this.move(issue, issue_above, issue_below);
     },
     move: function(issue, issue_above, issue_below){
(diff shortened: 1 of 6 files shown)


dahlbyk:
I've added `MIN_VALUE` and `MAX_VALUE` issues to https://huboard-rails-pr-314.herokuapp.com/huboard-testing/spooky-octo-pug that are handy to reproduce all sorts of edge case weirdness when you drag above/below them.

dahlbyk:
I also noticed that if you update the huboard metadata in an issue body, the Ember model (and thus the board) aren't updated accordingly.

discorick:
> I also noticed that if you update the huboard metadata in an issue body, the Ember model (and thus the board) aren't updated accordingly.

That is true, theres no compensating for it currently as no part of the client ever attempts to change that meta-data directly

dahlbyk:
> That is true, theres no compensating for it currently as no part of the client ever attempts to change that meta-data directly

But if someone does change the value while editing the description, it puts the board in an inconsistent state.

discorick:
> But if someone does change the value while editing the description, it puts the board in an inconsistent state.

Yeah, but if your hacking meta-data and things break...  ¯\_(ツ)_/¯ ?

dahlbyk:
> Yeah, but if your hacking meta-data and things break... ¯_(ツ)_/¯ ?

I empathize, but permanently breaking a board is no bueno.

ember-app/app/models/new/issue.js
```diff
@@ -187,7 +187,14 @@ var Issue = Model.extend({
...
     }.bind(this), "json");
-  }
+  },
+  maxMinOrderFix: function(){
+    var order = this.get('order');
+    while(order < 1e-319){ order *= 10; }
+    while(order > 1e307){ order /= 10; }
+
+    this.set('order', order);
+  }.on('init')
```
dahlbyk: Should this also refresh when the issue description changes, since our meta comment is editable?

ember-app/app/models/new/issue.js
```diff
@@ -187,7 +187,14 @@ var Issue = Model.extend({
       this.set("data.body_html", response.body_html);
       return this;
     }.bind(this), "json");
-  }
+  },
+  maxMinOrderFix: function(){
+    var order = this.get('order');
+    while(order < 1e-319){ order *= 10; }
```
dahlbyk: This loop will be infinite if `order <= 0` - should probably guard against that.

ember-app/tests/unit/mixins/cards/card-move-test.js
```diff
@@ -0,0 +1,81 @@
...
+var sut;
+
+function createIssueData(){
+  return Ember.Object.create({
+    _data: { order: 2.5 },
+    data: Ember.Object.create({ id: 123, title: 'test' })
+  });
+}
+
+function minMaxZeroOrderFixAlgorithm(issue){
```
dahlbyk: It strike me as odd to repeat this code here. Is there somewhere more accessible we could put it to reuse?

ember-app/tests/unit/mixins/cards/card-move-test.js
```diff
@@ -0,0 +1,81 @@
...
+moduleFor('mixin:cards/card-move', 'CardMoveMixin', {
+  setup: function(){
+    sut = Ember.Object.
+      createWithMixins(CardMoveMixin, {});
+  }
+});
+
+test('move the issue to the top', function(assert){
+  var cardMover = sut.cardMover;
+  cardMover.set('orderKey', 'order');
```
dahlbyk: In tests I like to set this sort of thing to a never-used-for-real value like `test_order` to make sure we're using `orderKey` everywhere rather than `"order"` on accident.

ember-app/tests/unit/mixins/cards/card-move-test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  //Moves issue to the top
+  var issueBelow = createIssueData();
+  var result = cardMover.moveToTop(issueBelow);
+  assert.ok(result < issueBelow._data.order, 'The new order is less than the issue below');
+
+  //Moves issue to the top when order has been adjusted for MIN_VALUE
+  issueBelow = createIssueData();
+  issueBelow._data.order = 5e-324;
```
dahlbyk: Use `Number.MIN_VALUE`?

ember-app/tests/unit/mixins/cards/card-move-test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  result = cardMover.moveToTop(issueBelow);
+  assert.ok(result < issueBelow._data.order, 'The new order is less than the issue below');
+
+  //If issue move is impossible give it a value below the minThreshold
+  issueBelow = createIssueData();
+  issueBelow._data.order = 1e-323;
+
+  result = cardMover.moveToTop(issueBelow);
+  assert.ok(result > cardMover.minThreshold, 'The new order is less than the issue below');
```
dahlbyk: Assert message does not match the scenario.

ember-app/tests/unit/mixins/cards/card-move-test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  result = cardMover.moveToBottom(issueAbove);
+  assert.ok(result > issueAbove._data.order, 'The adjusted order is greater than the issue below');
+
+  //If issue move is impossible give it a value above the maxThreshold
+  issueAbove = createIssueData();
+  issueAbove._data.order = 1.797692738796007e+308;
+
+  result = cardMover.moveToTop(issueAbove);
+  assert.ok(result > cardMover.minThreshold, 'The new order is less than the issue below');
```
dahlbyk: Assert message does not match the scenario.

ember-app/app/mixins/cards/card-move.js
```diff
@@ -11,20 +11,28 @@ var CardMoveMixin = Ember.Mixin.create({
...
       var above_order = issue_above._data[this.get("orderKey")];
       var below_order = issue_below._data[this.get("orderKey")];
       return (above_order + below_order) / 2;
     },
-    moveToTop: function(issue, issue_below){
-      return (issue_below._data[this.get("orderKey")]) / 2;
+    moveToTop: function(issue_below){
+      var below_order = issue_below._data[this.get("orderKey")]
+      var order = below_order / this.orderMultiplier;
+      while(order < this.minThreshold){ order *= 10; }
```
dahlbyk: This loop will be infinite if `order <= 0` - should probably guard against that.

ember-app/app/models/new/issue.js
discorick: Since the orders in the meta-data and the model are decoupled on the client it wouldn't make any difference in this case.

ember-app/app/models/new/issue.js
discorick: The zero order fix should make that impossible, but I agree it should be guarded.

ember-app/tests/unit/mixins/cards/card-move-test.js
discorick: Tests can be hard enough to read as it is, I typically try not to worry about reusability for object mock ups.

ember-app/tests/unit/mixins/cards/card-move-test.js
discorick: The card mover has a dependency on 'orderKey' equaling either `order` or `milestone_order`, are you suggesting giving the mockup issue data a 'test_order' key instead?

ember-app/app/models/new/issue.js
dahlbyk: So the larger question then is if we should refresh the meta-data when the description has been edited. It only affects people who damper with their cards, but our users _are_ developers.

ember-app/tests/unit/mixins/cards/card-move-test.js
dahlbyk: Mostly just worried about this implementation of the fix algorithm getting out of date with the real one.

ember-app/tests/unit/mixins/cards/card-move-test.js
dahlbyk: Yeah, just to reinforce that there's nothing special about `order` as a value.

ember-app/tests/unit/mixins/cards/card-move-test.js
discorick: Oh my eyes jumped to line 11, yeah I agree duping that logic is not great..

ember-app/app/models/new/issue.js
discorick: The parser mechanism could handle it easily for sure, I just don't see value in offering that kind of 'developer mode' control right now.

ember-app/app/mixins/cards/card-move.js
```diff
@@ -23,13 +23,16 @@ var CardMoveMixin = Ember.Mixin.create({
...
+      if(order <= 0){ order = issue_below.get('data.id') * this.minThreshold; }
       while(order < this.minThreshold){ order *= 10; }
 
       return order;
     },
     moveToBottom: function(issue_above){
       var above_order = issue_above._data[this.get("orderKey")];
       var order = above_order * this.orderMultiplier;
+
+      if(order === Infinity){ order = issue_above.get('data.id') / this.maxThreshold; }
```
dahlbyk: `issue_above.get('data.id') / this.maxThreshold` is going to give us a really small value, not a really big one.

Let's go with `this.maxThreshold / 1e10 * issue_above.get('data.id')`, which will turn `id = 160008709` into `order = 1.60008709e+305`. We'll run into problems if GitHub users create another 9.8 billion issues, which seems like enough breathing room; if you disagree, pick something bigger than 10. :smiling_imp:

dahlbyk:
Seems rock solid - nice work!

Issue #310, opened 2016-06-01
Label filters with the same name as a linked repo bind incorrectly


rauhryan:
## Steps to reproduce

I set up a test repo, https://huboard.com/huboard-testing/links-label-bug

The repo is linked to `huboard-testing/test` and also has a `test` label

Filter down to the `test` label and then copy paste the URL into a new tab, it will re-produce the bug

![](http://f.cl.ly/items/3504142F3D200J3N1p06/Screen%20Recording%202016-06-01%20at%2012.49%20PM.gif)

PR #316, opened 2016-06-06
Prevents QP's from polluting others with matching name

discorick:
### Problem:

An example would be a label named 'pizza' and a repo named 'pizza' - a url with query parameters to filter on the label pizza would also trigger the repo 'pizza' filter. 

### Solution:

The Query Parameters now only apply themselves to the filter group they belong to.

fixes https://github.com/huboard/huboard-web/issues/310


dahlbyk:
Nothing to add about the code, and behavior seems correct to me:
https://huboard-rails-pr-316.herokuapp.com/huboard-testing/links-label-bug#/?label=["test"]

:shipit:

PR #320, opened 2016-06-08
Adds deployed boilerplate for PR timelines

discorick:
PR's throw an exception in development when there is a deployment requested event in the timeline.


rauhryan:
Ah, there must not be a template for that event type.

dahlbyk:
Seems simple enough... :shipit:?

Issue #195, opened 2016-02-02
Enhancing the Review Process


dylanfitz:
I work with a team of about 10 people, and often we need someone to review an issue before it can get moved to "done". Currently, we just place it in review and "@" mention the person who needs to review that item, however, with a growing team that is managing much larger projects, this is becoming difficult to manage. 

We are looking for a way to add team members to an issue that are responsible for reviewing the work done in that issue. Some key information that would be helpful is if the reviewer has opened the issue since it was placed in review, and maybe an indicator that the user approves this work. It would also be helpful if these reviewers received a notification email when an issue is moved into "Review". It is important that the review is tied to a user, and not just a free floating feature. It is also important that more than one reviewer can be added to an issue.

dahlbyk:
This is very similar to https://github.com/huboard/huboard/issues/612 - what do you think of https://github.com/huboard/huboard/issues/612#issuecomment-175121205?

seanlinsley:
@dahlbyk I think it's important to assign the PR itself to the reviewer from inside Github instead of storing this information as metadata.

dahlbyk:
> I think it's important to assign the PR itself to the reviewer from inside Github instead of storing this information as metadata.

I tend to agree, we're just limited to one GitHub-provided assignee slot. The idea would be to use metadata to support tracking a secondary set of assignees that HuBoard would let you swap into the GH Assignee field as appropriate for your team's process.

seanlinsley:
What I had in mind was that from Huboard there could only be two assignees: the implementer and the reviewer. The implementer would continue to work as it currently does, but the reviewer would be stored as metadata until a PR was created, at which point the PR itself would be assigned to the reviewer.

My concerns with doing any more than that to fit various team processes are:
- the assignee(s) wouldn't be visible from Github
- the feature would take a lot longer to build, and maintain
- it would be harder for users to understand the feature, and configure it how they want

dahlbyk:
> the reviewer would be stored as metadata until a PR was created, at which point the PR itself would be assigned to the reviewer.

Interesting...so you'd want to set the expected reviewer on the _issue_ so assignments can be automated on the PR. That hadn't occurred to me - I am intrigued. I'm not sure that we're completely set up to inspect PRs' referenced issues to track down a reviewer, but it's definitely an interesting idea.

[...]

PR #323, opened 2016-06-15
Multiple assignee support

discorick:
Enables the multiple assignees API with backwards compatibility for Enterprise

Relies on https://github.com/huboard/ghee/pull/43
Closes https://github.com/huboard/huboard/issues/612
Related to #195

### DIFF
--- a/app/controllers/api/issues_controller.rb
+++ b/app/controllers/api/issues_controller.rb
@@ -127,16 +127,18 @@ module Api
 
     def assign_issue
       user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
-      @issue = huboard.board(user, repo).issue(number)
-        .add_assignees(assignees)
+      issue = huboard.board(user, repo).issue(number)
+      @issue = issue.add_assignees(assignees) if issue['assignees']
+      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees'] 
       @assignee = assignees.first
       render json: @issue
     end
 
     def unassign_issue
       user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
-      @issue = huboard.board(user, repo).issue(number)
-        .remove_assignees(assignees)
+      issue = huboard.board(user, repo).issue(number)
+      @issue = issue.remove_assignees(assignees) if issue['assignees']
+      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees'] 
       @assignee = assignees.first
       render json: @issue
     end
--- a/app/controllers/api/issues_controller.rb
+++ b/app/controllers/api/issues_controller.rb
@@ -126,9 +126,16 @@ module Api
     end
 
     def assign_issue
-      user, repo, number, @assignee = params[:user], params[:repo], params[:number], params[:assignee]
+      user, repo, number, @assignees = params[:user], params[:repo], params[:number], params[:assignees]
       @issue = huboard.board(user, repo).issue(number)
-        .patch 'assignee' => @assignee
+        .add_assignees(@assignees)
+      render json: @issue
+    end
+
+    def unassign_issue
+      user, repo, number, @assignees = params[:user], params[:repo], params[:number], params[:assignees]
+      @issue = huboard.board(user, repo).issue(number)
+        .remove_assignees(@assignees)
       render json: @issue
     end
 
--- a/app/controllers/api/issues_controller.rb
+++ b/app/controllers/api/issues_controller.rb
@@ -128,18 +128,24 @@ module Api
     def assign_issue
       user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
       issue = huboard.board(user, repo).issue(number)
-      @issue = issue.add_assignees(assignees) if issue['assignees']
-      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees'] 
       @assignee = assignees.first
+      if issue['assignees']
+        @issue = issue.add_assignees(assignees)
+      else
+        @issue = issue.patch('assignee' => @assignee)
+      end
       render json: @issue
     end
 
     def unassign_issue
       user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
       issue = huboard.board(user, repo).issue(number)
-      @issue = issue.remove_assignees(assignees) if issue['assignees']
-      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees'] 
       @assignee = assignees.first
+      if issue['assignees']
+        @issue = issue.remove_assignees(assignees)
+      else
+        @issue = issue.patch('assignee' => @assignee)
+      end
       render json: @issue
     end
 
--- a/app/controllers/api/issues_controller.rb
+++ b/app/controllers/api/issues_controller.rb
@@ -126,16 +126,18 @@ module Api
     end
 
     def assign_issue
-      user, repo, number, @assignees = params[:user], params[:repo], params[:number], params[:assignees]
+      user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
       @issue = huboard.board(user, repo).issue(number)
-        .add_assignees(@assignees)
+        .add_assignees(assignees)
+      @assignee = assignees.first
       render json: @issue
     end
 
     def unassign_issue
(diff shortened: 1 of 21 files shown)


app/controllers/api/issues_controller.rb
```diff
@@ -126,9 +126,20 @@ def archive_issue
     end
 
     def assign_issue
-      user, repo, number, @assignee = params[:user], params[:repo], params[:number], params[:assignee]
-      @issue = huboard.board(user, repo).issue(number)
-        .patch 'assignee' => @assignee
+      user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
+      issue = huboard.board(user, repo).issue(number)
+      @issue = issue.add_assignees(assignees) if issue['assignees']
+      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees']
```
dahlbyk: Maybe not idiomatic Ruby, but I'd prefer an explicit `if ... else ...` block rather than a pair of statements with inverted conditions. Seems like it would be easier to see exactly what's happening, since we need to do one or the other.

app/controllers/api/issues_controller.rb
```diff
@@ -126,9 +126,20 @@ def archive_issue
...
+      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees'] 
+      @assignee = assignees.first
+      render json: @issue
+    end
+
+    def unassign_issue
+      user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
+      issue = huboard.board(user, repo).issue(number)
+      @issue = issue.remove_assignees(assignees) if issue['assignees']
+      @issue = issue.patch('assignee' => assignees.first) if !issue['assignees']
```
dahlbyk: Another spot I would suggest `if ... else ...`.

app/controllers/api/issues_controller.rb
dahlbyk: Also, not sure how much effort has historically been put into avoiding breaking clients between a push and a refresh, but it seems like it would be relatively easy to be defensive in this case:

```ruby
user, repo, number, assignees = params[:user], params[:repo], params[:number], params[:assignees]
@assignee = params[:assignee] || @assignees.first
if issue['assignees']
  @issue = issue.add_assignees(assignees)
else
  @issue = issue.patch('assignee' => @assignee)
end
render json: @issue
```

ember-app/app/models/new/issue.js
```diff
@@ -102,14 +102,33 @@ var Issue = Model.extend({
...
+      _self.set("assignee", response.assignee);
+      return _self;
+    });
+  },
+  unassignUsers: function(logins){
+    var _self = this;
+    return Ember.$.post(`${this.get("apiUrl")}/unassigncard`, {
+      assignees: logins, 
+      correlationId: this.get("correlationId")
+  }, function(){}, "json").then(function( response ){
```
dahlbyk: Missing indentation here

ember-app/app/models/new/issue.js
```diff
@@ -102,14 +102,33 @@ var Issue = Model.extend({
...
     return Ember.$.post(`${this.get("apiUrl")}/assigncard`, {
-      assignee: login, 
+      assignees: logins, 
       correlationId: this.get("correlationId")
     }, function(){}, "json").then(function( response ){
-      this.set("assignee", response.assignee);
-      return this;
-    }.bind(this));
+      var assignees = response.assignees;
+      if(assignees && assignees.any((assignee)=>{ assignee.login === logins.first })){
```
dahlbyk: Use `anyBy` here and below, as we do elsewhere?

dahlbyk: And `logins.first` seems like it shouldn't work... `logins.first()`?

dahlbyk:
Noticed uses of both `anyBy` and `isAny`...the former has been deprecated in favor of the latter since https://github.com/emberjs/ember.js/pull/3662.

ember-app/app/mixins/subscriptions/board.js
```diff
@@ -98,19 +98,21 @@ var BoardSubscriptionMixin = Ember.Mixin.create({
...
-        copy = `${actor} assigned #${message.issue.number} to ${assignee.login}`;
-      } else {
-        copy = `${actor} unassigned #${message.issue.number}`;
-      }
+      var copy = `${actor} assigned #${message.issue.number} to ${assignee}`;
+      this.get("flashMessages").info(copy);
+    }, {time: 5000}),
+    issueUnassigned: sortedQueue(function(message){
+      var actor = message.actor.login;
+      var assignee = message.assignee ? message.assignee : "a member";
```
dahlbyk: `message.assignee || "a member"`

Maybe "someone" instead of "a member"?

ember-app/app/mixins/subscriptions/card.js
```diff
@@ -32,14 +32,40 @@ var CardSubscriptionMixin = Ember.Mixin.create({
...
-    },
+    assigned: sortedQueue(function(message){
+      var assignees = this.get("issue.assignees");
+      if(assignees && !assignees.isAny("login", message.assignee.login)){
+        if(message.assignee.login){
+          this.get("issue.assignees").pushObject(message.assignee);
+        } else {
+          var assignee = this.get("issue.repo.assignees").find((a)=>{
+            return a.login === message.assignee;
+          });
```
dahlbyk: Can use `findBy` here.

ember-app/app/components/hb-multi-assignee.js
```diff
@@ -0,0 +1,93 @@
...
+      event.preventDefault();
+      event.stopPropagation();  
+      this.set("isOpen", false);
+    }.bind(this));
+
+    this.$(".close").on('click.flyout', function(event){
+      if(Ember.$(event.target).is("[data-ember-action],[data-toggle]")){return;}
+      if(Ember.$(event.target).parents("[data-ember-action],[data-toggle]").length){return;}
+      this.set("isOpen", false);
+    }.bind(this));
```
dahlbyk: Isn't the `function(item) { ... }.bind(this)` pattern unnecessary if you use `(item) => { ... }`?

ember-app/app/components/hb-multi-assignee.js
```diff
@@ -0,0 +1,93 @@
...
+      } else {
+        this.$().removeClass("open");
+      }
+    },
+    assignTo: function(assignee) {
+      var action;
+      var selected = this.get('selected');
+      if(selected.isAny('login', assignee.login)){
+        action = 'unassign'
+        var obj = selected.find((select)=>{ return select.login === assignee.login });
```
dahlbyk: `findBy` here, too!

app/controllers/api/issues_controller.rb
discorick: Re `if ... else ..` no problem, I wouldn't say whats there is idiomatic ruby (probably closer to an anti-pattern :P )

> but it seems like it would be relatively easy to be defensive in this case

Sorry, what are we defending against?

ember-app/app/components/hb-multi-assignee.js
discorick: I don't think so, the context of `this` doesn't change using the shorthand function notation (or whatever the official term for it is)

app/controllers/api/issues_controller.rb
dahlbyk: Someone with the old JS will submit `assignee` rather than `assignees`.

ember-app/app/components/hb-multi-assignee.js
dahlbyk: "Arrow function expressions" have ["lexical `this`"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Lexical_this), which is explicitly stated as an alternative to a "bound function".

dahlbyk:
> Noticed uses of both anyBy and isAny...the former has been deprecated in favor of the latter since emberjs/ember.js#3662.

Can we swap out the uses of `anyBy`?

discorick:
> Can we swap out the uses of anyBy?

Yep, I sed'd the whole app.

dahlbyk:
I can't seem to break this. :shipit:?

N.B. showing multiple assignees on cards will be part of https://github.com/huboard/huboard/issues/661.

discorick:
I can't break it either... 🚢  if it feels good!

PR #325, opened 2016-06-20
Implements a github repository image uploader storage class for carrierwave

rauhryan:
Closes huboard/huboard#520 when merged

Todo:

  - [x] add :org/:repo support for the uploader
  - [x] make sure we only allow fast forward commits
  - [x] disable uploads if logged in user does not have push bit
  - [x] Change both the upload end-points to POST
  - [x] optimize upload paths with the folder structure convention 

Later: 

  - [x] reach out to enterprise users for how to handle no push bit (fork or fall-back user)
  - [x] reach out to GitHub to see how they feel about the approach


config/routes.rb
```diff
@@ -46,7 +46,8 @@
 
 
   namespace :api do
-    get 'uploads/asset' => 'uploads#asset_uploader'
+    get  'uploads/asset' => 'uploads#asset_uploader'
```
dahlbyk: Unrelated to this PR, but this should almost certainly be a `POST`.

lib/carrier_wave/storage/git_hub.rb
```diff
@@ -0,0 +1,94 @@
+module CarrierWave
+  module Storage
+    class GitHub < File
+
+      def store!(file)
+        upload_path = "#{SecureRandom.uuid}/#{file.original_filename}"
```
dahlbyk: We should partition the root tree so it doesn't get huge. I'm inclined to copy how Git stores its object database:

```ruby
blob = repo.git.blobs.create(...)
uploadPath = "#{blob[0,2]}/#{blob[2,38]}/#{file.original_filename}"
```

lib/carrier_wave/storage/git_hub.rb
```diff
@@ -0,0 +1,94 @@
...
+          commit = repo.git.commits.create({ tree: tree["sha"], message: "uploading #{file.original_filename}" })
+
+          #create a ref
+          ref = repo.git.refs.create({
+            ref: "refs/huboard/uploads",
+            sha: commit["sha"]
+          })
+        else
+          parent_commit = ref.body["object"]["sha"]
+          base_tree = repo.git.trees(parent_commit)
```
dahlbyk: Does this work? I would expect you'd want to use `repo.git.commits(parent_commit).tree`.

lib/carrier_wave/storage/git_hub.rb
```diff
@@ -0,0 +1,94 @@
...
+          }],
+          base_tree: base_tree["sha"]})
+
+          #create a new commit
+          commit = repo.git.commits.create({ tree: tree["sha"], message: "uploading #{file.original_filename}" , parents: [parent_commit]})
+
+          #update the ref
+          ref = repo.git.refs("huboard/uploads").patch({
+            sha: commit["sha"]
+          })
```
dahlbyk: We'll need to test/handle this not being a fast-forward update (meaning another upload was pushed since we fetched the ref).

dahlbyk:
So if you want to pull down the uploads ref, update your remote config with a new `fetch` spec:

```
[remote "origin"]
        url = https://github.com/rauhryan/skipping.stones.git
        fetch = +refs/heads/*:refs/remotes/origin/*
        fetch = +refs/huboard/*:refs/remotes/origin/hb/*
```

lib/carrier_wave/storage/git_hub.rb
dahlbyk: Nevermind, it looks like this is working fine. Git/GitHub must be resolving the commit SHA to its tree. :cool:

dahlbyk:
To fill in some details from offline discussion...

> * add :org/:repo support for the uploader
> * Change both the upload end-points to POST

The idea here is to replace `get  'uploads/asset'` with a route like `post /:user/:repo/upload` that will return appropriate upload "instructions". For `AwsUploader`, the SaaS behavior, the instructions allow `POST`ing directly to AWS. For the new GitHub uploader, we need `ApiUploader#direct_fog_url` to return `/:user/:repo/upload/create` or equivalent so we can upload into the correct repository.

> * make sure we only allow fast forward commits

`force: false` is the default [ref update](https://developer.github.com/v3/git/refs/#update-a-reference) behavior, so really we just need to make sure we can retry if the `PATCH` is rejected.

> * optimize upload paths with the folder structure convention

See https://github.com/huboard/huboard-web/pull/325#discussion_r67751456

> * reach out to enterprise users for how to handle no push bit (fork or fall-back user)

The downside of a fork is that the assets would be deleted if the fork is deleted. I really don't like the risk of data loss. My preference would be to simply disallow read-only user uploads unless a fall-back user token has been specified in the appliance configuration.

rauhryan:
@discorick @dahlbyk I'm pretty sure I've addresses all the concerns, can we merge this?

dahlbyk:
> I'm pretty sure I've addresses all the concerns, can we merge this?

Sorry, haven't taken the time to review. Will try to do that tonight.

PR #326, opened 2016-06-22
Disables ajax caching in Ember

discorick:
Some browsers (specifically FF) use cached ajax responses when the browser restores tabs by default. This can cause chaos when cards start getting moved that are not actually there, and just a general poor experience. 

Relates to the conversation happening here: https://github.com/huboard/huboard/issues/677

### DIFF
--- a/ember-app/app/app.js
+++ b/ember-app/app/app.js
@@ -3,6 +3,7 @@ import Resolver from 'ember/resolver';
 import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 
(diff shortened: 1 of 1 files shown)


dahlbyk:
Does FF really use cached responses even if the response says not to? It seems preferable to adjust response headers if they're not properly signaling to Firefox (or other browsers) that the content is not reusable, but if FF doesn't respect such headers...

discorick:
I think playing with the headers is a better and more effective approach for sure. I just don't have the bandwidth to work on this further until next month. Is this a good enough solution in your estimation ?

PR #328, opened 2016-06-23
Implements multiple assignees for new issues as well

discorick:
Allows multiple assignees to be added to a new issue

### DIFF
--- a/ember-app/app/controllers/issue/create.js
+++ b/ember-app/app/controllers/issue/create.js
@@ -18,6 +18,10 @@ var IssuesCreateController = Ember.Controller.extend({
   }.observes('selectedRepo'),
   otherLabels: Ember.computed.alias("selectedRepo.data.other_labels"),
   assignees: Ember.computed.alias("selectedRepo.data.assignees"),
+  //Ugly Hack for backwards Enterprise Compatibiltiy
+  assigneesEnabled: function(){
+    return this.get("selectedRepo.board.issues").isAny("assignees");
+  }.property("selectedRepo.board.issues"),
   milestones: Ember.computed.alias("selectedRepo.data.milestones"),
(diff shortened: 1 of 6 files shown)


ember-app/app/controllers/issue/create.js
```diff
@@ -18,6 +18,10 @@ var IssuesCreateController = Ember.Controller.extend({
   }.observes('selectedRepo'),
   otherLabels: Ember.computed.alias("selectedRepo.data.other_labels"),
   assignees: Ember.computed.alias("selectedRepo.data.assignees"),
+  //Ugly Hack for backwards Enterprise Compatibiltiy
+  assigneesEnabled: function(){
+    return this.get("selectedRepo.board.issues").isAny("assignees");
```
dahlbyk: This seems likely to be slow on large repos without `assignees` support. Can we just look at the first one?

dahlbyk:
Looks good other than the concern about perf when assignees isn't supported. I'd still love to be able to configure the review app to enable/disable the assignees preview to test each scenario.

ember-app/app/controllers/issue/create.js
discorick: The base implementation for `any` short-circuits on `found` https://github.com/emberjs/ember.js/blob/v2.6.0/packages/ember-runtime/lib/mixins/enumerable.js#L662

dahlbyk: Right. It's efficient when `assignees` is expected to be found, but will scan every issue looking for `assignees` when it's not found.

discorick: No bueno on this, GitHub API returns issues with any assignees without an assignees key :(

discorick:
Alight added a guard for old client code here: 99cf61a

Issue #288, opened 2016-05-05
Continuous Integration with Travis CI


dahlbyk:
If tests aren't automated, are they really passing?

* [x] Run Ruby tests
* [x] Run `ember-app/` tests
* [x] Run `ember-accounts/` tests

https://docs.travis-ci.com/user/getting-started/
https://docs.travis-ci.com/user/languages/ruby
https://docs.travis-ci.com/user/languages/javascript-with-nodejs

dahlbyk:
> Run Ruby tests

As of #305, building PRs plus `master` and `enterprise` branches.

PR #333, opened 2016-07-04
Adds the ember app build to travis

discorick:
Runs ember tests for CI
Fixes all the failing tests, CI is now green!

Closes https://github.com/huboard/huboard-web/issues/288


dahlbyk:
> Add rake build 6710cc8efc316af54e1ec6d571300cc8d28214e5

Build output showed the Rake build wasn't actually running. Commands copied directly from previous Ruby builds.

`<sarcasm>`It's awesome how long it takes to restore dependencies!`</sarcasm>`

dahlbyk:
`npm install` is stalling, but I'm expecting this to be green for all three builds.

dahlbyk:
Also, the jshint hang is a known issue (https://github.com/ember-cli/ember-cli/issues/1083) that's apparently fixed by a newer version of testum (https://github.com/testem/testem/issues/528).

discorick:
I think your hanging npm install is due to: http://status.npmjs.org/incidents/nl2mpgltpnvj

dahlbyk:
> I think your hanging npm install is due to: http://status.npmjs.org/incidents/nl2mpgltpnvj

:+1: good find.

> Travis optimizations

Stumbled on the Ember apps' own Travis config files and pulled in some of their seemingly useful bits (specifically caching).

dahlbyk:
@discorick ember-accounts build is still timing out after NPM issue cleared up. Thoughts?

discorick:
> @discorick ember-accounts build is still timing out after NPM issue cleared up. Thoughts?

I can reproduce in development, using `ember-cli 0.2.7`, The model unit tests are probably what the problem is as per the issue you referenced. We could either using different ember cli versions for each test, or move the `ember-app` to a current build of ember-cli

discorick:
@dahlbyk been mulling on this. I originally omitted the accounts app for a couple of reasons:
  - It should not really be a part of this repo
  - We will not be accepting open source contributions for it

Given those factors, I am content not including it into CI. What do you think ?

dahlbyk:
> Given those factors, I am content not including it into CI. What do you think ?

I'm fine with that for now; better to get CI in for the rest.

Two more lint errors now with ember-accounts removed.

dahlbyk:
> Use phantom 2.0

Fascinating.... I assume you borrowed that config from somewhere. Link?

discorick:
Uses phantom 2.0 as per: https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-177592725

PR #327, opened 2016-06-22
Card redesign layout

discorick:
Pathway to 🚢 
  - [x] Task card layouts
  - [x] Issue Blocked/Ready labels as states
  - [x] Milestone card layouts
  - [x] Issue Reference UX
     - [x] also https://github.com/huboard/huboard-web/pull/338#issuecomment-237854443
  - [x] Scrolling task columns
  - [x] Acceptable Clientside Perf

closes: https://github.com/huboard/huboard/issues/261
closes: https://github.com/huboard/huboard-web/issues/318
closes: https://github.com/huboard/huboard/issues/372

PR #335, opened 2016-07-07
Treat blocked/ready labels as states

discorick:
Merges https://github.com/huboard/huboard-web/pull/327  (the card design branch.)

Blocked/Ready labels now act readily as states on HuBoard. If the label is missing <s>it is created</s> the health check fails. 

closes https://github.com/huboard/huboard/issues/399

### DIFF
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -154,12 +154,12 @@ var Issue = Model.extend({
   }.observes('stateLabel'),
   customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", {
     get:function(){
-      var state = this.get("_data.custom_state");
+      var state = this.get("stateLabel");
       if(state){ return state; }
-      return this.get('stateLabel');
+      return this.get("_data.custom_state");
     },
     set: function (key, value) {
-      var previousState = this.get("_data.custom_state") || this.get("stateLabel");
+      var previousState = this.get("stateLabel") || this.get("_data.custom_state");
       this.set("_data.custom_state", value);
 
       var endpoint = value === "" ? previousState : value;
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -184,6 +184,7 @@ var Issue = Model.extend({
       this.set("_data.custom_state", value);
 
       var endpoint = value === "" ? previousState : value;
+      if(!endpoint){ return; }
       var options = {
         dataType: "json",
         data: {correlationId: this.get("correlationId")},
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -167,17 +167,14 @@ var Issue = Model.extend({
         return name === 'blocked' || name === 'ready';
     }) || '';
   }.property('data.other_labels.[]'),
-  stateLabelObserver: function(){
-    if(this.get('stateLabel')){this.set('_data.custom_state', this.get('stateLabel'));}
-  }.observes('stateLabel'),
-  customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", {
+  customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", "stateLabel", {
     get:function(){
       var state = this.get("stateLabel");
       if(state){ return state; }
       return this.get("_data.custom_state");
     },
     set: function (key, value) {
-      var previousState = this.get("stateLabel") || this.get("_data.custom_state");
+      var previousState = this.get("_data.custom_state") || this.get("stateLabel");
       this.set("_data.custom_state", value);
 
       var endpoint = value === "" ? previousState : value;
@@ -192,9 +189,9 @@ var Issue = Model.extend({
       Ember.$.ajax(options)
       .then(function(response){
         this.set("processing", false);
-        this.set("other_labels", response.other_labels);
         this.set("data.body", response.body);
         this.set("data.body_html", response.body_html);
+        this.set("other_labels", response.other_labels);
       }.bind(this));
 
       return value;
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -168,7 +168,7 @@ var Issue = Model.extend({
     }) || '';
   }.property('data.other_labels.[]'),
   stateLabelObserver: function(){
-    this.set('_data.custom_state', this.get('stateLabel'));
+    if(this.get('stateLabel')){this.set('_data.custom_state', this.get('stateLabel'));}
   }.observes('stateLabel'),
   customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", {
     get:function(){
--- a/ember-app/app/models/new/issue.js
+++ b/ember-app/app/models/new/issue.js
@@ -82,6 +82,9 @@ var Issue = Model.extend({
     var changedColumns = this.get("data.current_state.index") !== column.data.index;
     if(changedColumns){ this.set("data._data.custom_state", ""); }
 
+    var state_label = this.get('other_labels').find((l)=> { return this.isStateLabel(l)});
+    this.get('other_labels').removeObject(state_label);
+
     this.set("data.current_state", column.data);
     this.set("data._data.order", index);
     return Ember.$.post(`${this.get("apiUrl")}/dragcard`, {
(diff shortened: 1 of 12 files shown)


dahlbyk:
> For discussion sake you can cause strange behavior by:
> 
> - Assigning both a blocked/ready label to an issue, this will enable both states at once
> 
> The algorithm isn't overly bright, it will just pick the first state label and run with it. Personally, I am comfortable with this - when wrestling with outside influences (like label changes from GitHub) theres only so much we can do to guard against non-sensical states.

I think it reasonable for us to assign reasonable priority to our labels. If something is Blocked, it is explicitly not Ready. Once un-Blocked, it's reasonable to immediately fall back into a Ready state. My $.02.



> If the label is missing it is created.

I'm unsure about this, since many boards already have some sort of "impediment" label that they might want to just rename for use by HuBoard (ideally we'd make the Blocked label configurable - `huboard.yml` strikes again!).

What if we added a Block/Ready label checker to the health check? "[Click Here] to create, or rename your existing labels to..."

Nice-to-have: option in health check to apply the labels to currently blocked/ready-without-label cards.

discorick:
> I think it reasonable for us to assign reasonable priority to our labels. If something is Blocked, it is explicitly not Ready. Once un-Blocked, it's reasonable to immediately fall back into a Ready state. My $.02.

I suppose before each label update on the client we can make sure only "one or the other" before we make any PUT's. The desired state will be the first in the in the array.

> I'm unsure about this, since many boards already have some sort of "impediment" label that they might want to just rename for use by HuBoard (ideally we'd make the Blocked label configurable - huboard.yml strikes again!).

Until we improve our UX around how we communicate features, the health check route will likely just push this feature into the dark along with other less obvious HB features. I propose revisit this once we have yml config rolled out.

dahlbyk:
OK, so...thoughts on how to communicate either:

1) You don't have permission to create labels, so your Ready/Block state couldn't be set with a label, or
2) We just created a label for you, so you might want to flip into GitHub, delete it and rename your existing Impediment label to match HB's expectation.

------

> Until we improve our UX around how we communicate features, the health check route will likely just push this feature into the dark along with other less obvious HB features.

We should add analytics to identify how many boards are "unhealthy", and how often the health page is used.

discorick:
> 1) You don't have permission to create labels, so your Ready/Block state couldn't be set with a label

Being able to label/unlabel an issue means you are a collaborator, implicitly you also can create them. 

> 2) We just created a label for you, so you might want to flip into GitHub, delete it and rename your existing Impediment label to match HB's expectation.

Flash message: "HuBoard just created the blocked label for you"

dahlbyk:
> Being able to label/unlabel an issue means you are a collaborator, implicitly you also can create them.

Touché

> Flash message: "HuBoard just created the blocked label for you"

That works. Sticky? Can we add a link to FAQ to suggest renaming an existing label, if they have one?

discorick:
> That works. Sticky? Can we add a link to FAQ to suggest renaming an existing label, if they have one?

Seems good to me.

> Once un-Blocked, it's reasonable to immediately fall back into a Ready state.

Won't be too bad to handle in real-time on the client. The effort required to sort out how to handle an issue with both blocked and ready labels from GitHub seems more costly than its worth. Like I said before, it's just nonsensical to begin with.

dahlbyk:
> Like I said before, it's just nonsensical to begin with.

It is, but deterministic handling doesn't seem prohibitively expensive. The same is true of our column labels - are we nondeterministic, or is there a rule (e.g. highest number wins)?

discorick:
> It is, but deterministic handling doesn't seem prohibitively expensive. The same is true of our column labels - are we nondeterministic, or is there a rule (e.g. highest number wins)?

Well, It's already deterministically choosing blocked state if both labels are there, I just don't think we should remove the counterpart label automatically via the API.

discorick:
@dahlbyk, starting to warm up to your point of view about automatically creating labels. I don't love the UX of it. We can start by just treating labels that are 'blocked' or 'ready' as states, then later figure out a good way to treat them as a health check ? (Or maybe in this PR)

discorick:
As per @dahlbyk , should the label picker deselect 'blocked' labels if 'ready' is selected or vice versa ?

ember-app/app/models/new/issue.js
```diff
@@ -142,24 +169,18 @@ var Issue = Model.extend({
...
   },
   stateLabel: function(){
-    var state_label = this.get('other_labels').map((label)=> {
-        return label.name.toLowerCase();
-      }).find((name)=>{
-        return name.toLowerCase() === 'blocked' || name.toLowerCase() === 'ready';
-    });
-    return state_label || "";
+    return this.get('other_labels').map((label)=>{return label.name.toLowerCase()}).find((name)=>{
+      return name === 'blocked' || name === 'ready';
```
dahlbyk: Could this just be `this.get('other_labels').find(this.isStateLabel)`?

ember-app/app/models/new/issue.js
```diff
@@ -142,24 +169,18 @@ var Issue = Model.extend({
...
-  customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", {
+  customState: Ember.computed("data._data.custom_state", "data.other_labels.[]", "stateLabel", {
     get:function(){
       var state = this.get("stateLabel");
       if(state){ return state; }
       return this.get("_data.custom_state");
     },
     set: function (key, value) {
-      var previousState = this.get("stateLabel") || this.get("_data.custom_state");
+      var previousState =  this.get("stateLabel") || this.get("_data.custom_state");
```
dahlbyk: :trollface:

lib/health_checking/health_checks/board/issue_blocked_label_check.rb
```diff
@@ -0,0 +1,31 @@
+module HealthChecking
+  module HealthChecks
+    module Board
+      class IssueBlockedLabelCheck
+        include HealthCheck
+        name 'issue_blocked_label_check'
+        weight :warning
+        authorization :collaborator
+        passed "Issue blocked label is present"
+        failed "Issue blocked label is not present"
```
dahlbyk: Maybe `"Blocked" label is not present`? "Issue" seems redundant, and maybe more explicitly mentioning the expected label name would nudge people toward renaming their own? Or we could add some sort of tooltip to give a help bubble with more explanation.

dahlbyk:
Other than clarifying the health check for those who might already have equivalent-ish labels, I think we're pretty good here.

> should the label picker deselect 'blocked' labels if 'ready' is selected or vice versa ?

I feel pretty strongly that using the label picker directly should behave like a pure label picker. If you happen to pick both Ready and Blocked (or pick Blocked while Ready is already selected, or whatever), then our state UX should respect the label with the highest precedence (Blocked) but keep labels as-selected.

dahlbyk:
Two observations from just a bit more testing:

- Moving a blocked/ready card between columns is changing state in the description, but not removing the label.
- Not new to this PR, but it seems like Blocked state should only be removed if the card is moved to the right on the board. I could see a blocked card in Next Release moving back into Backlog while still blocked.

ember-app/app/models/new/issue.js
discorick: Labels don't have their own their own intelligent models, so theres not really an efficient way other than checking every single label on its creation and setting the bool (as opposed to a computed property).

ember-app/app/models/new/issue.js
discorick: 😲

lib/health_checking/health_checks/board/issue_blocked_label_check.rb
discorick: Yeah, that copy is better, issue is redundant. 

> Or we could add some sort of tooltip to give a help bubble with more explanation.

I think for our UX 'funnel' phase we should identify things like this and determine a good way to provide intuitive info about HuBoard on the fly

discorick:
> Moving a blocked/ready card between columns is changing state in the description, but not removing the label.

Shoot! That _should_ be working, I will look into it. 

> Not new to this PR, but it seems like Blocked state should only be removed if the card is moved to the right on the board. I could see a blocked card in Next Release moving back into Backlog while still blocked.

Hmm, thats probably true in many cases (where the issue remains blocked). Do we want to tackle this here though?

ember-app/app/models/new/issue.js
dahlbyk: Oh, I mean passing `this.isStateLabel` as a function as the `filter` callback. `isStateLabel` doesn't use `this` at all, it's a proper function (input :point_right: output).

dahlbyk:
> Do we want to tackle this here though?

Probably not, just thought I'd mention it while it was top of mind.

discorick:
> Moving a blocked/ready card between columns is changing state in the description, but not removing the label.

Found the problem, sneaky edge case good catch.

dahlbyk:
:tada:

PR #342, opened 2016-07-19
Fixes cases where new issues can duplicate orders, making them dissapear

### LINKED ISSUE #341: Card disappearing on reorder
Ran into a case where a care disappears on every move, then reappears  when a render is forced on on the milestones view:

uhoh

The card below meta-data (test-issue#7)
image

The card above meta-data (test-issue#6)
image

The meta-data of the card being moved:
image

discorick:
closes #341

### DIFF
--- a/ember-app/app/mixins/cards/card-move.js
+++ b/ember-app/app/mixins/cards/card-move.js
@@ -20,8 +20,9 @@ var CardMoveMixin = Ember.Mixin.create({
       var below_order = issue_below._data[this.get("orderKey")];
       return (above_order + below_order) / 2;
     },
-    moveToTop: function(issue_below){
-      var below_order = issue_below._data[this.get("orderKey")];
+    moveToTop: function(issue_below, key){
+      key = key ? key : this.get("orderKey");
+      var below_order = issue_below._data[key];
       var order = below_order / this.orderMultiplier;
 
       if(order <= 0){ order = issue_below.get('data.id') * this.minThreshold; }
@@ -29,8 +30,9 @@ var CardMoveMixin = Ember.Mixin.create({
 
       return order;
     },
-    moveToBottom: function(issue_above){
-      var above_order = issue_above._data[this.get("orderKey")];
+    moveToBottom: function(issue_above, key){
+      key = key ? key : this.get("orderKey");
+      var above_order = issue_above._data[key];
       var order = above_order * this.orderMultiplier;
 
       if(order === Infinity){ order = this.maxThreshold / 1e10 * issue_above.get('data.id'); }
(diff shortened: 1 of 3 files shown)


ember-app/app/mixins/cards/card-move.js
```diff
@@ -20,17 +20,19 @@ var CardMoveMixin = Ember.Mixin.create({
       var below_order = issue_below._data[this.get("orderKey")];
       return (above_order + below_order) / 2;
     },
-    moveToTop: function(issue_below){
-      var below_order = issue_below._data[this.get("orderKey")];
+    moveToTop: function(issue_below, key){
+      key = key ? key : this.get("orderKey");
```
dahlbyk: `key || this.get("orderKey")`

That said, can we just kill the `orderKey` property altogether? This implict statefulness seems a likely source of other bugs.

discorick: I am fairly certain we can. 👍

dahlbyk:
To clarify my understanding, what's the expected behavior on Milestones when reordering from Tasks, and vice versa?

ember-app/app/mixins/cards/card-move.js
discorick: Actually.. not so straight-forward - I just remembered why its there in the first place . It's not impossible to acheive but will probably cause more changes than I would hope for in this PR.

discorick:
>  To clarify my understanding, what's the expected behavior on Milestones when reordering from Tasks, and vice versa?

Are you wanting to know the expected result from the creation of a new issue in both respective views? Or what is happening every time a card is moved.

dahlbyk:
> Are you wanting to know the expected result from the creation of a new issue in both respective views? Or what is happening every time a card is moved.

Whatever is relevant to this PR? :grinning: 

To clarify, this regression was introduced in https://github.com/huboard/huboard-web/pull/314 when `cardMover.moveToTop()` was introduced with new usage patterns. The fix seems sound.

discorick:
So, on the task view, when an issue is created the following happens:
  - The order of the top issue in the column is taken (if there is one) and the new issue's `order` key is set
  - Then, the column is resorted based on `milestone_order`, and the new issues `milestone_order` key is set based on the top issue
  - Finally we post and the issue is created. 

This is the same for issues created in the milestone view. The reason this bug was easy to miss is the module used to calculate the order of the issue would automatically choose the key (`milestone_order, or `order`) based on the current view, causing one of the calculations to be incorrect based on the view.

dahlbyk:
Actual milestone order change looks good, but I'm sometimes seeing two issues rendering:
double-issue

dahlbyk:
Behavior confirmed :+1:

PR #343, opened 2016-07-20
[WIP] Include the url referrer in server-side page jobs

### LINKED ISSUE #313: Analytics correlation is still not working correctly
#254 doesn't seem to be working as we hoped it would. We should ping Segment support to see if there's any way to fix this.

discorick:
I suspect our goal tracking is incomplete because server side `page()` calls for analytics don't include referral urls. 

Fixes #313


discorick:
Turns out this solution will not work :( . The segment lib actually pulls in `analytics.js` from GA which in turn stores a cookie. The cookie is how goals know how to correlate page calls.

PR #344, opened 2016-07-21
Moves goal tracking page calls to the client

discorick:
Adds click handlers to provide "meta-page" calls for GA goal tracking. Closes https://github.com/huboard/huboard-web/issues/313.


discorick:
image

app/views/shared/_analytics.html.erb
```diff
@@ -2,6 +2,17 @@
 <% if ENV['SEGMENTIO_KEY'] %>
   window.analytics=window.analytics||[],window.analytics.methods=["identify","group","track","page","pageview","alias","ready","on","once","off","trackLink","trackForm","trackClick","trackSubmit"],window.analytics.factory=function(t){return function(){var a=Array.prototype.slice.call(arguments);return a.unshift(t),window.analytics.push(a),window.analytics}};for(var i=0;i<window.analytics.methods.length;i++){var key=window.analytics.methods[i];window.analytics[key]=window.analytics.factory(key)}window.analytics.load=function(t){if(!document.getElementById("analytics-js")){var a=document.createElement("script");a.type="text/javascript",a.id="analytics-js",a.async=!0,a.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.io/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(a,n)}},window.analytics.SNIPPET_VERSION="2.0.9",
   window.analytics.load("<%= ENV['SEGMENTIO_KEY'] %>");
+  var $ = $ ? $ : false;
+  if($){
+    $('a').click(function(e){
```
dahlbyk: If we don't qualify this selector, I would expect some internal links to be double-tracked.

dahlbyk:
I'm fuzzy how the Authenticated event is firing since there's round-trip to GitHub involved. What am I missing?

discorick:
Oh, we don't really need it anymore. `/dashboard` is the new authenticated. Does that seem reasonable ?

dahlbyk:
> Oh, we don't really need it anymore. `/dashboard` is the new authenticated. Does that seem reasonable ?

:+1:

app/views/shared/_analytics.html.erb
discorick: Watching on testing, I didn't notice any dups, we can qualify though, same as the buttons

PR #345, opened 2016-07-25
Filters ux improvements

discorick:
Various tweaks and improvements to the filters panel:

- More compact filters
- Filter groups are now collapsible
- Filter groups now have active filters counts
- Collapsed/Uncollapsed is saved into localStorage

image
image

### DIFF
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -593,6 +593,7 @@
     margin: 0px;
     h5 {
       padding: 2px 0px 2px 4px;
+      cursor: pointer;
       &:hover {
         background: #efefef;
         border-color: #efefef;
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -578,12 +578,38 @@
 
 .filters li a {
   position:relative;
-  padding: 10px;
+  padding: 4px;
+  font-size: 12px;
   display:block;
   padding-right:25px;
   text-decoration: none;
   border-radius: 4px;
   color:#777;
+}
+
+.filters {
+  .filter-group {
+    h5 {
+      &:hover {
+        background: #efefef;
+        border-color: #efefef;
+        border-style: solid;
+        border-radius: 2px;
+        border-width: 0px;
+      }
+      i {
+        position: relative;
+        float: right;
+        padding: 2px 2px 0px 0px;
+        font-size: 14px;
+      }
+    }
+    &.collapsed {
+      .filter {
+        display: none;
+      }
+    }
+  }
 }
 
 .filters li a strong { 
--- a/app/assets/stylesheets/_board.scss
+++ b/app/assets/stylesheets/_board.scss
@@ -589,7 +589,10 @@
 
 .filters {
   .filter-group {
+    list-style: none;
+    margin: 0px;
     h5 {
+      padding: 2px 0px 2px 4px;
       &:hover {
         background: #efefef;
         border-color: #efefef;
--- a/ember-app/app/components/hb-filter-group.js
+++ b/ember-app/app/components/hb-filter-group.js
@@ -1,0 +1,12 @@
+import Ember from 'ember';
+
+var HbFilterGroupComponent = Ember.Component.extend({
+  classNames: ['filter-group'],
+  classNameBindings: ['isCollapsed:collapsed'],
+  isCollapsed: true,
+  click: function(){
+    this.toggleProperty('isCollapsed');
+  }
+});
+
+export default HbFilterGroupComponent;
--- a/ember-app/app/components/hb-filter-group.js
+++ b/ember-app/app/components/hb-filter-group.js
@@ -4,11 +4,16 @@ var HbFilterGroupComponent = Ember.Component.extend({
   tagName: ['ul'],
   classNames: ['filter-group'],
   classNameBindings: ['isCollapsed:collapsed'],
-  isCollapsed: false,
+  setCollapsedKey: function(){
+    if(this.get(`settings.filters${this.get('name')}Collapsed`)){
+      this.set(`settings.filters${this.get('name')}Collapsed`, false);
+    }
+  }.on('init'),
+  isCollapsed: Ember.computer.alias(`settings.filters${this.get('name')}Collapsed`),
   click: function(el){
     if(this.$(el.target).is('.filter-header') || this.$(el.target).is('.ui-icon')){
-      this.toggleProperty('isCollapsed');
-      //save to LS
+      this.toggleProperty(`settings.filters${this.get('name')}Collapsed`);
+      this.notifyPropertyChange('isCollapsed');
     }
   }
 });
--- a/ember-app/app/components/hb-filter-group.js
+++ b/ember-app/app/components/hb-filter-group.js
@@ -5,17 +5,22 @@ var HbFilterGroupComponent = Ember.Component.extend({
   classNames: ['filter-group'],
   classNameBindings: ['isCollapsed:collapsed'],
   setCollapsedKey: function(){
-    if(this.get(`settings.filters${this.get('name')}Collapsed`)){
+    if(!this.get(`settings.filters${this.get('name')}Collapsed`)){
       this.set(`settings.filters${this.get('name')}Collapsed`, false);
     }
   }.on('init'),
-  isCollapsed: Ember.computer.alias(`settings.filters${this.get('name')}Collapsed`),
+  isCollapsed: function(){
+    return this.get(`settings.filters${this.get('name')}Collapsed`);
+  }.property(''),
   click: function(el){
     if(this.$(el.target).is('.filter-header') || this.$(el.target).is('.ui-icon')){
       this.toggleProperty(`settings.filters${this.get('name')}Collapsed`);
       this.notifyPropertyChange('isCollapsed');
     }
-  }
+  },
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/_board.scss
```diff
@@ -578,14 +578,40 @@
...
+.filters {
+  .filter-group {
+    h5 {
+      &:hover {
+        background: #efefef;
+        border-color: #efefef;
+        border-style: solid;
+        border-radius: 2px;
+        border-width: 0px;
+      }
```
dahlbyk: Can we add a `cursor` style here?
firefox_2016-07-28_13-52-44

ember-app/app/templates/filters.hbs
```diff
@@ -1,25 +1,31 @@
 <ul class='filters'>
-  <h5>Repos</h5>
-  {{#each filters.boardFilters as |filter|}}
-    {{hb-filter name=filter.name mode=filter.mode color=filter.color tagType="colored" queryParam="repo"}}
-  {{/each}}
+  {{#hb-filter-group name='Repos'}}
```
dahlbyk: Can we hide this filter if there aren't any linked repos?

ember-app/app/templates/filters.hbs
```diff
@@ -1,25 +1,31 @@
...
 
-  <h5>Milestones</h5>
-  {{#each filters.milestoneFilters as |filter|}}
-    {{hb-filter name=filter.name mode=filter.mode color=filter.color queryParam="milestone"}}
-  {{/each}}
-  <h5>Labels</h5>
-  {{#each filters.labelFilters as |filter|}}
-    {{hb-filter name=filter.name mode=filter.mode color=filter.color tagType="colored" queryParam="label"}}
-  {{/each}}
+  {{#hb-filter-group name='Milestones'}}
```
dahlbyk: Can we hide this group if it's just the "No milestone" filter?

ember-app/app/templates/filters.hbs
```diff
@@ -1,25 +1,31 @@
...
+    {{#each filters.milestoneFilters as |filter|}}
+      {{hb-filter name=filter.name mode=filter.mode color=filter.color queryParam="milestone"}}
+    {{/each}}
+  {{/hb-filter-group}}
+
+  {{#hb-filter-group name='Labels'}}
```
dahlbyk: Can we hide this group if there are no labels?
https://huboard-rails-pr-345.herokuapp.com/huboard-testing/empty

dahlbyk:
While we're fiddling with filters, thoughts on [this idea](https://github.com/huboard/huboard/issues/683#issuecomment-232857565)?

> Under "Active members" is a reasonable place to expect the other assignment-based filters. We could move them up to a single line below the avatars: me • others • unassigned.

(Maybe format them button-ish instead of using a delimiter?)

As to the general change:

* :+1: for less vertical space.
* I feel like there's not enough left padding on the non-label filter links.
* Nor is there enough padding on the label group headers. Alternative here would be to drop the hover background color, which I think would be fine (with a `cursor` change instead).
* Not sure if it's worth making Cards expandable. What if the assignment proposal were applied here, merging three lines into one: Issues • Pull Requests?
* I'm feeling like expanded may be a better default, since that what folks are used to. Let's see what feedback we get on collapsibility before adjusting what's visible by default?

Other random (potential) scope creep:

* Can we push an event into Segment (and then on to GA) on expand or collapse? My hypothesis is that only Milestones is likely to be collapsed to show more labels; otherwise I doubt the toggle will be used.
* It seems like expanded/collapsed groups should persist between Tasks and Milestones. Currently switching back and forth resets all to collapsed every time.
* Maybe even push expand/collapse state into local storage to persist between sessions for the given repo?

ember-app/app/templates/filters.hbs
discorick: Could? With the introduction of stateful labels, this seems like a feature for nobody. Our app will persistently nag people until they have blocked/ready labels present.

ember-app/app/templates/filters.hbs
discorick: 👍 Like it!

ember-app/app/templates/filters.hbs
discorick: 👍 Like it!

app/assets/stylesheets/_board.scss
discorick: Good call, pointer would be better!

discorick:
👍  Moving up assignees closer to the avatar filters
👍  More work with the padding
👍  Expanded may be a better default - Given we also store collapsed/uncollapsed state in localstorage

👎  Segment. I mean, I would like to do it but we need to implement tracking on the ember-app first 😄 

> Not sure if it's worth making Cards expandable. What if the assignment proposal were applied here, merging three lines into one: Issues • Pull Requests?

It's probably outside the scope of this one. Need roll a whole one-off component for card filters to make this possible, which touches query parameters etc.

ember-app/app/templates/filters.hbs
dahlbyk: Good point, hadn't considered the state labels. :+1: to leave it.

dahlbyk:
> > Not sure if it's worth making Cards expandable. What if the assignment proposal were applied here, merging three lines into one: Issues • Pull Requests?
> 
> It's probably outside the scope of this one. Need to roll a whole one-off component for card filters to make this possible, which touches query parameters etc.

Help me understand this. You mean we would need a component different from `hb-filter` to style them differently? Could we just make that filter accept a `class` (or apply a `class` to the list's parent) to apply a different layout?

Also, I just noticed the filter markup looks like:

```html
<ul class="filters">
  <div class="filter-group">
    <h5 class="filter-header">Group</h5>
    <li class="filter">Uno</li>
    <li class="filter">Dos</li>
    ...
  </div>
  <div class="filter-group">...</div>
  ...
</ul>
```

Can we split the parent `<ul>` into a `<ul>` per group?

```html
<div class="filter-group">
  <h5>Group</h5>
  <ul class="filters">
    <li class="filter">Uno</li>
    <li class="filter">Dos</li>
    ...
  </ul>
</div>
<div class="filter-group">...</div>
...
```

ember-app/app/components/hb-filter-group.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+var HbFilterGroupComponent = Ember.Component.extend({
+  classNames: ['filter-group'],
+  classNameBindings: ['isCollapsed:collapsed'],
+  isCollapsed: true,
+  click: function(el){
+    if(el.target.className === 'filter-header'){ this.toggleProperty('isCollapsed'); }
```
dahlbyk: Could this check `nodeName === 'h5'` rather than an otherwise unused `class`?

discorick: Doesn't using `nodeName` as the selector seem more easily broken? Any future work would more than likely preserve the class structure, even if DOM elements get swapped was my original thought. I could be wrong though.

discorick:
> Can we split the parent <ul> into a <ul> per group?

Glad you said something, I was already thinking about the same thing!

rauhryan:
Playing around with it I have two thoughts

1. I like the slimmed down padding and font-size 👍  we have done a great job of slimming the cards down the filters feel like a positive step 

2. The behavior of the collapsed state doesn't feel right when there are active filters. There is nothing to let me know there are active filters other than our clear filters button in the toolbar. For example if I add a filter and then refresh the board, the default state is the filter groups are collapsed. The user needs to open each filter group to figure out which filters are active. It "feels" unfinished in a way 

perhaps the collapsed state lists the active filters? perhaps we have some kinds of (1) <- number of active filters in the collapsed header?

ember-app/app/components/hb-filter-group.js
dahlbyk: That's fair, should we decide to switch `h5` to `h4 or something. Carry on.

dahlbyk:
> perhaps the collapsed state lists the active filters?

:+1:

discorick:
> The behavior of the collapsed state doesn't feel right when there are active filters. There is nothing to let me know there are active filters other than our clear filters button in the toolbar. For example if I add a filter and then refresh the board, the default state is the filter groups are collapsed. The user needs to open each filter group to figure out which filters are active. It "feels" unfinished in a way

👍 , Just pushed active filter counts for each group

> Maybe even push expand/collapse state into local storage to persist between sessions for the given repo?

Done and done!

ember-app/app/templates/components/hb-filter-group.hbs
```diff
@@ -1,5 +1,10 @@
 <h5 class='filter-header'>
   {{name}}
+  {{#if activeFilters.length}}
+    <small>({{activeFilters.length}})</small>
+  {{/if}}
   <i class="ui-icon {{if isCollapsed 'ui-icon-triangle-1-e' 'ui-icon-triangle-1-s'}}"></i>
 </h5>
-{{yield}}
+{{#each filters as |filter|}}
+  {{hb-filter name=filter.name mode=filter.mode color=filter.color tagType="colored" queryParam="label"}}
```
dahlbyk: Will need to get `tagType` and `queryParam` from the parent component.

discorick: The `tagType` I did omit on accident. `queryParam` is deprecated for that component, just never got taken out.

dahlbyk:
:tada: :exclamation:

PR #347, opened 2016-07-28
Replaces Intercom with Gitter Sidecar in the app

discorick:
Playing with using Gitter (wired to `huboard/huboard`) rather than intercom to help facilitate more thought sharing and discussion in the app.

### DIFF
--- a/app/views/layouts/ember.html.erb
+++ b/app/views/layouts/ember.html.erb
@@ -14,13 +14,13 @@
   </head>
   <body class="flex-body" id="application">
 
-     <%= render partial: "shared/analytics" %>
      <script>
       window.EMOJIS = <%= emojis.to_json.html_safe %>;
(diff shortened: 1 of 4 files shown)


app/views/layouts/ember.html.erb
```diff
@@ -14,13 +14,12 @@
   </head>
   <body class="flex-body" id="application">
 
-     <%= render partial: "shared/analytics" %>
```
dahlbyk: I don't think we want to remove this, just disable Intercom in Segment.

discorick: Can do that as well, though for the ember app I believe the only thing we use it for is the intercom chat widget.

dahlbyk: 98% sure we also feed GA via Segment.

discorick: Oh right for the board page you on GA you are correct.

PR #349, opened 2016-08-02
Enables ga funnel for trials

discorick:
Uses the `data-analytics-path` convention to send client-side pages to segment on trial activation


vendor/engines/saas/app/views/saas/trials/show.html.erb
```diff
@@ -4,7 +4,7 @@
     <div class="trial_form">
       <%= form_tag("/settings/profile/#{@user}/trial/activate", method: :post) do %>
         <input type="email" placeholder="Email Address" name="billing_email" required>
-        <input class="btn primary" type="submit" value="Do it!"/>
+        <input class="btn primary" type="submit" value="Do it!" data-analytics-path="/settings/<%=@user%>/trial/activate"/>
```
dahlbyk: Should this be `/settings/#{owner}/trialing` rather than `/trial/activate`? Or are you planning to update the funnel?

discorick: I am going to adjust the funnel, wanted to match the actual route - seems to be simpler ?

dahlbyk: :+1:

dahlbyk: On further consideration, using a different URL will let us differentiate in GA between old and new tracking.

discorick: Also a bonus 👍

PR #350, opened 2016-08-03
Fix copyright dates

discorick:
Calculates copyright based on `Time.now.year`


dahlbyk:
I believe _technically_ notices are supposed to included the start year as well:

```
Copyright &copy; 2014&ndash;<%= Time.now.year %> HuBoard, Inc.
```

discorick:
@rauhryan is 2014 the correct copyright start year ?

PR #351, opened 2016-08-03
Hides create issue card from anonymous users

### LINKED ISSUE #180: Add Issue should not be offered to anonymous users

discorick:
Previously, anonymous users would be allowed to attempt issue creation but fail ultimately. 

fixes #180


dahlbyk:
:clap:

rauhryan:
//TODO: move that App global to a service or something

PR #359, opened 2016-08-10
[Experimental] Adds smoke-and-mirrors addon for perf

discorick:
Adds the the smoke-and-mirrors ember addon to test improvements to render performance. 

Problems:
  - Dependency https://github.com/runspired/ember-run-raf has an override for the global `clearTimeout`, unfortunately that messes with `Faye.ENV.clearTimeout` . `Faye.ENV` is actually the app's global scope, and when `clearTimeout` is called, the overridden version of the method is invoked. This in turn throws an `illegal invocation` error on the browser. I have the error suppressed for the time being by wiping out the `clearTimeout` method on `Faye.ENV` 

- Attempting to utilize `vertical-collection` resulted in error  @rauhryan is looking into it


discorick:
I was playing with https://github.com/DockYard/ember-in-viewport as well. Using scrolling columns is going to require quite a bit of sophistication to get card elements hiding/appearing properly. A Naive implementation here: https://ilos.video/pAoLub

discorick:
Update:

Continuing work on the `smoke-and-mirrors` library today, was unable to reach a reasonable level of success rendering cards. The closest I was able to get was adding an observer on on the `vertical-collection` component class itself (monkey patch), along with force rendering every item. This is both a hack, and breaks columns ability to scroll. 

@rauhryan I think I am just about ready to call this one a no go, unless you would like to take another stab ?

rauhryan:
I'd love to pull out the ember-cli upgrade commits or at least get that done on master. Otherwise, I think this PR can be closed

discorick:
Hm, I might open up a PR against `discorick/card-design-layouts` with 92ecf01 to prevent any merge conflicts, does that sound ok ?

PR #360, opened 2016-08-17
Scrollable Columns Performance

discorick:
Adds a `ScrollableColumnMixin` which fires events for scrolling based on a simple tolerance. Initial POC implementation reveals additional cards during scroll and hides them again on scroll up. 

Greatly improves initial performance on Firefox. 

TODO:
  - [x] Accurate Scrolling Logic
  - [x] Filters
  - [x] Events that add/remove issues
  - [x] Creating new issues
  
Known bugs:
  <s>- Scrolling really quickly can throw the list out of sync</s>
  <s>- Scrolling upwards while dragging issues breaks if that issue is removed from the array</s>

### DIFF
--- a/ember-app/app/components/columns/hb-column.js
+++ b/ember-app/app/components/columns/hb-column.js
@@ -113,7 +113,7 @@ var HbColumnComponent = Ember.Component.extend(SortableMixin, ScrollingColumn, {
 
     var issuesLength = this.get('visibleIssues').length;
     if(horizonVisible && issuesLength < totalColumnLength){
-      Ember.run.debounce(this, 'revealIssues' , horizon, 200);
+      Ember.run.throttle(this, 'revealIssues' , horizon, 200);
     }
   }.on('columnScrolledDown'),
   revealIssues: function(horizon){
--- a/ember-app/app/components/columns/hb-column.js
+++ b/ember-app/app/components/columns/hb-column.js
@@ -136,11 +136,12 @@ var HbColumnComponent = Ember.Component.extend(SortableMixin, ScrollingColumn, {
     }
   }.on('columnScrolledUp'),
   hideIssue: function(){
+    if(this.get('isHovering')){ return; }
+
     var lastItem =  this.get('visibleIssues.lastObject');
     lastItem = this.get('visibleIssues').indexOf(lastItem);
     this.get('visibleIssues').removeAt(lastItem);
     this.set('cardIndex', lastItem);
-    this.$('.cards').superSortable('refresh');
   }
 });
 
--- a/ember-app/app/components/columns/hb-column.js
+++ b/ember-app/app/components/columns/hb-column.js
@@ -10,7 +10,8 @@ var HbColumnComponent = Ember.Component.extend(SortableMixin, ScrollingColumn, {
 
   columns: Ember.computed.alias("model.columns"),
   visibleIssues: function(){
-    return this.get('sortedIssues').slice(0, 10);
+    var index = this.get('cardIndex') + 10;
+    return this.get('sortedIssues').slice(0, index);
   }.property('sortedIssues.[]'),
   sortedIssues: function(){
     return this.get("model.sortedIssues");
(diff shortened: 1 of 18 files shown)


discorick:
@rauhryan here is the scrollbar bug I am running into: 

scrollbarbug

rauhryan:
Yeah that's not a bug, that's how OSX does scrollbars :)

discorick:
Alright things are looking pretty tight. The one bug I have identified is attempting to scroll and item straight to the bottom of the list doesn't work if the list is scrolled - this is a known [jquery bug](https://bugs.jqueryui.com/ticket/9314) and is fixed in version `1.12.0`, we are currently using `1.9.1` for the sortable library. 

@rauhryan is it safe to replace our jquery build ? I would just swap it with the newest but I doubt we need all the addons, do you know what we need?

rauhryan:
@discorick it should be safe to upgrade us to the latest jquery-ui but I'm almost positive we should be on 1.11.4, I swear I upgraded it when I was working on superSortable.... but maybe not


I can give it a shot real quick

discorick:
@rauhryan  That would be awesome, yeah we just need `1.12.x` to get the scroll fixes in there.

rauhryan:
@discorick done

discorick:
@rauhryan thanks! Unfortunately the upgrade didn't fix as promise :( 

Scrolling a card from the top down to the bottom still dissallows dragging beneath the lowest card.

rauhryan:
It doesn't seem consistent, are sure it's jquery

discorick:
Not sure no, just seemed the most likely explanation. Hmm this is another tough one.

ember-app/app/mixins/cards/sortable.js
```diff
@@ -13,6 +13,8 @@ var SortableMixin = Ember.Mixin.create(CardMoveMixin, {
     var newSortable = _self.get('newSortable');
     var sortableFx = newSortable ? "superSortable" : "sortable";
     this.$(".cards")[sortableFx]({
+      containment: "document",
+      tolerance: "pointer",
```
rauhryan: I actually disagree with pointer for the cards, I don't think it's better 

 > "intersect": The item overlaps the other item by at least 50%.

The behavior should be based on the bounding box of the card not the pointer of the mouse. 

I would agree that the autoscrolling could use a little bit of work to be "better" but that should be in the superSortable stuff, not the mouse tolerance

dahlbyk:
> short curcuit scrolling if filters are active

Why do we need to do this?

discorick:
> > short curcuit scrolling if filters are active

> Why do we need to do this?

For the scrolling optimization, we maintain a seperate array that keeps popping in new items  (which in turn render to the DOM) as the list scrolls downward.

This is troublesome with filters because we are not able to determine if an item is filtered or not currently unless it is in the DOM - in other words the filter set will only filter on items already on the DOM, and doesn't know whether the next batch of DOM elements should be filtered or not .

rauhryan:
> short curcuit scrolling if filters are active
> Why do we need to do this?

Because the scrolling handlers push and remove items from visibleIssues with is a computed array

In theory, computed properties should be immutable, but I can see in this case where it gets weird figuring out what properties to observe and compute based off 

An alternated would be to recompute the visibleIssues array based on the cardIndex property, but then the list refresh calls have to become some kind of observer based on who knows what and it becomes hard to follow 

TL;DR; two-way binding is hard... rendering things is hard...

discorick:
> An alternated would be to recompute the visibleIssues array based on the cardIndex property, but then the list refresh calls have to become some kind of observer based on who knows what and it becomes hard to follow

Even in the case of filters, cards would need to be added to the DOM before we could tell if they matched the filters

ember-app/app/components/columns/hb-column.js
```diff
@@ -113,7 +113,7 @@ var HbColumnComponent = Ember.Component.extend(SortableMixin, ScrollingColumn, {
 
     var issuesLength = this.get('visibleIssues').length;
     if(horizonVisible && issuesLength < totalColumnLength){
-      Ember.run.debounce(this, 'revealIssues' , horizon, 250);
+      Ember.run.debounce(this, 'revealIssues' , horizon, 200);
```
rauhryan: I'd love to see what this feels like as a throttle

I think throttle is more appropriate in this situation, because you want the event to fire right away just not too many calls in a row, where as debounce as to wait until you've scrolled all the way to the end and then it pauses if you scroll really fast.

discorick: I though about that too, its worth experimenting with. My concern with is the first one fires immediately, but the scroll is so fast the next one is throttled and the user will have to scroll back up and down again to trigger it.

PR #362, opened 2016-08-22
Updates to the latest warden

rauhryan:


discorick:
Is this just an update to the `minitest` gem? I don't see any other changes.

Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: git://github.com/huboard/warden-github.git
-  revision: 459903681fa00820065348c27639b9f27db9fa55
+  revision: 3c3cf6943b01ff12cfe22a1ad1202ddcc4721e66
```
rauhryan: ^ updates our warden-github gem to the lastest git sha on https://github.com/huboard/warden-github

rauhryan:
> Is this just an update to the minitest gem? I don't see any other changes.

not sure why bundler wanted to update minitest, but the most important change is the sha change for the huboard/warden-github gem

Gemfile.lock
discorick: Oh interesting, just for my own learning here, does that mean when looking for the warden gem, it grabs that remote and that sha on every (gem) install ?

rauhryan: Yep, we have a dependency on so custom code in the warden-github gem to figure out what scopes we have requested. 

The Gemfile.lock specifies that we want that exact sha. 

So, our heroku build will pull that new sha on the next deploy

PR #363, opened 2016-08-29
Column layout with FF overrides

rauhryan:
Rauhryan/card design layout+ff hacks

### DIFF
--- a/ember-app/app/mixins/cards/sortable.js
+++ b/ember-app/app/mixins/cards/sortable.js
@@ -3,8 +3,6 @@ import CardMoveMixin from "../cards/card-move";
 
 
 var SortableMixin = Ember.Mixin.create(CardMoveMixin, {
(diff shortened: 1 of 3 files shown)


ember-app/app/mixins/cards/sortable.js
```diff
@@ -21,20 +19,16 @@ var SortableMixin = Ember.Mixin.create(CardMoveMixin, {
...
+        clone.width(ui.outerWidth() - 8);
+        return clone;
       },
       start: function(ev, ui){
         ui.placeholder.height(ui.helper.outerHeight());
       },
       items: ".is-draggable",
       placeholder: "ui-sortable-placeholder",
       connectWith: ".cards",
-      over: function(){
```
discorick: This will kill the hover shadow for moving columns won't it ?

rauhryan: yes, we don't need it anymore  because of no scrolling :)

PR #358, opened 2016-08-09
Fixes infinite loop create board loop

discorick:
Fixes the infinite loop and closes #357 

Additionally, as the changes required some styling, applied a bit of a facelift to the board create process altogether:

When prompted to create a board (buttons light up on hover):
image

When third party access restrictions are present:
image

When push bit is missing:
image

@rauhryan @dahlbyk What do you think? Improvement ?

PR #364, opened 2016-08-31
Refactors to BEM and cleans up some markup

rauhryan:
Relates to #358 

I jumped into the comments on #358 and wanted to help out. 

I started off by finding a [rough closed div tag](https://github.com/huboard/huboard-web/blob/591292cf52d8f74f0c04ddab7b120d0dad1b2605/app/views/board/create_board.html.erb#L44) inside the form tags and started backtracking and folding up all the markup to find out how to fix it. I then needed to add indenting so I'm very sorry for the diff not being very useful. 

Things I have address in the PR 

  - [Fixes href quoting problem](https://github.com/huboard/huboard-web/pull/358#discussion-diff-74652033)
  - [Check if the repo owner is an organization](https://github.com/huboard/huboard-web/pull/358#discussion_r74652878)
  - [Adds analytic paths to determine retrys or abandons](https://github.com/huboard/huboard-web/pull/358#discussion_r74652243)

### DIFF
--- a/app/assets/stylesheets/board/create.scss
+++ b/app/assets/stylesheets/board/create.scss
@@ -1,52 +1,34 @@
-.create-new-board--container {
+.create-new-board__container {
   max-width: 600px;
   margin-left: auto;
   margin-right: auto;
-}
-
-.create-new-board--message {
-  margin: 0px auto 30px auto;
-  margin-bottom: 30px;
+  padding-bottom: 30px;
   font-size: 15px;
 }
 
-.create-new-board--list {
+.create-new-board__message {
+  margin: 0px auto;
+}
+
+.create-new-board__list {
   list-style: none;
   padding-left: 10px;
-  padding-bottom: 10px;
   text-align: left;
-  li {
+  &__item {
     border-left-style: solid;
     border-left-color: $lightGrey;
     padding-left: 10px;
     margin-bottom: 14px;
-  }
-}
-
-.create-new-board--form {
-  input.btn.primary {
-    background: none;
-    font-size: 16px;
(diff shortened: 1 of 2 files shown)


discorick:
I am curious if you have a GA goal workflow in mind for abandons, we can infer them currently by any traffic that doesn't meet the goal requirements. Having said that 👍  for more data :D

discorick:
Not a fan of the big purple buttons ? 
image

image

rauhryan:
Idk based on the current implementation

you'd see 

  - /board/create => /board/creating
  - /board/create => /board/creating/abandon
  - /board/create => /board/creating => /board/creating/retry
  - /board/create => /board/creating => /board/creating/abandon

Or some infinity combination of /creating => (retry...abandon)

rauhryan:
> Not a fan of the big purple buttons ? 

It wasn't the design that threw me off, I actually prefered the bigger purple buttons. 

The css debt didn't seem worth it, there was a triple specific selector on a 'input.btn.primary' nested under another class 

There was also use of forms and form inputs inside anchor tags. They didn't really need to be forms and serve just fine as anchor tags that are styled.

I got carried away fixing the markup that I just simplified it all to use the built in bootstrap styles.

discorick:
> It wasn't the design that threw me off, I actually prefered the bigger purple buttons.

> The css debt didn't seem worth it, there was a triple specific selector on a 'input.btn.primary' nested under another class

Not sure I 100% follow - if we prefer a certain styling, why can't we achieve without the debt ? Aren't we optimizing in the opposite direction of where we want to go then?

PR #365, opened 2016-09-01
[DEMO] optimize filters a bit by filtering the sorted array first

rauhryan:


discorick:
Ah interesting snag here:

- Apply a filter
- Drag a card down until you know its passed some "hidden" cards
- Drop it in between a place you know some filtered cards exist
- The Above/Below Order of the card calculated based on whats in the DOM already and the card will fly up to whatever position it ends up in :(

PR #369, opened 2016-09-14
Early access support for Project boards in HuBoard

rauhryan:


dahlbyk:
Stuff that doesn't work:

- Sync should refresh the current project's cards
- Filtering to only Issue/PR doesn't work
- Filtering by repo doesn't work (and really shouldn't be an option)
- Filtering always includes Notes; the reverse should probably be true

PR #378, opened 2016-09-20
Adds the repo name beside the fullscreen issue title

discorick:
image

closes https://github.com/huboard/huboard/issues/537

### DIFF
--- a/app/assets/stylesheets/_issue.scss
+++ b/app/assets/stylesheets/_issue.scss
@@ -67,6 +67,16 @@
     line-height: 1em;
     font-size: 12px;
   }
+
+  .linked-repo {
+    padding-left: 4px;
+    font-size: 14px;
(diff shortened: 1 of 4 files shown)


dahlbyk:
Added a commit with equivalent padding between title and number.

#### Before
firefox_2016-09-21_10-36-32

#### After


-------

> Not completely sure, but I think the font-size on .ui-icon-note will be inherited from .linked-repo

`font-size` from `h2` wins.
firefox_2016-09-21_10-38-20

rauhryan:
> Not completely sure, but I think the font-size on .ui-icon-note will be inherited from .linked-repo
> font-size from h2 wins.

CSS you so silly

discorick:
Yeah, I am actually not sure why I added another property rather than using `model.repoName`, I will take care of it!

PR #383, opened 2016-10-03
Ember cli upgrade 2.4.3

rauhryan:


dahlbyk:
Glad to see you picked up on the failing build.

Let's hold off on PRs for additional ember-cli upgrades until we get this green, merged and deployed to production.

.travis.yml
```diff
@@ -34,4 +36,4 @@ install:
 
 script: 
   - bundle exec rake
-  - cd ember-app && pwd && ./node_modules/.bin/ember --version && cd ..
+  - cd ember-app && pwd && ./node_modules/.bin/ember test && cd ..
```
dahlbyk: Drop `&& pwd`?

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: 
   - ruby
   - node_js
+node_js:
+  - 4
```
dahlbyk: Is this needed?

ember-app/.travis.yml
```diff
@@ -0,0 +1,22 @@
+---
+language: node_js
+node_js:
+  - "4"
```
dahlbyk: Is this file actually needed?

dahlbyk:
> `ember-app/.travis.yml`
> 
> Is this file actually needed?

Eh? (I only ask because it's brand new.)

--------

Other than that I think we're good here.

rauhryan:
> ember-app/.travis.yml
> 
> Is this file actually needed?
> Eh? (I only ask because it's brand new.)

Answer:
Ember-cli keeps adding/expecting it to be there. The upgrade process keeps adding from the `ember init` command. 

It's nice to have it there because `ember init` will give me a diff of what the default travis yml is supposed to look like, which helps to know if there is some change that needs to be ported to our actual travis yml

PR #384, opened 2016-10-03
Ember cli upgrade 2.4.3

rauhryan:


discorick:
Is this ready to merge into its parent or are you still working on it ?

rauhryan:
it's done, @dahlbyk wanted there to be deployable steps at each point release

rauhryan:
not exactly what I was going for here, perhaps there is a better way to communicate the intent

2.3 should have merge first, followed by this one

dahlbyk:
Yeah, PR into PR probably isn't the right flow here - at this point #383 has turned into a 2.4.3 upgrade when it comes into `master`. Let's continue the conversation there.

PR #385, opened 2016-10-18
Configurable expanded labels on cards

rauhryan:


ember-app/app/initializers/punch-component.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
```
discorick: Not familiar with the "punch" verbiage, curious where it comes from?

discorick:
I like the setting for expanded labels being global, it helps keep the visualization of cards consistent across boards. The only potential downside is how it expands milestones as well... I wonder if users will still want an expanded view for the taskboard, but more compact on milestones or vice versa.

discorick:
image

Minor, PR icon is missing purple styling.

rauhryan:
> I like the setting for expanded labels being global, it helps keep the visualization of cards consistent across boards. The only potential downside is how it expands milestones as well... I wonder if users will still want an expanded view for the taskboard, but more compact on milestones or vice versa.

We can ship and see if anyone gives that feedback

rauhryan:
I've addressed all feedback, so I'm going to merge

PR #387, opened 2016-10-28
adds a milestoned event

rauhryan:
captures the milestoned event and publishes it to the client as an edited event

https://developer.github.com/changes/2016-10-27-new-label-and-milestone-webhooks/

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -75,12 +75,12 @@ module ApplicationHelper
   end
   def generate_issue_event(action, message)
     verb = action.present_tense
-    constant = "Api::Issues#{verb.capitalize}IssueJob".constantize
+    job = JobResolver.find_event("issue", verb)
 
     if Rails.configuration.active_job.queue_adapter == :sucker_punch # sucker_punch doesn't support enqueue
-      constant.perform_later message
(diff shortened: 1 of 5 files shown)


dahlbyk:
We also need to handle `demilestoned`.

I noticed this new job is identical to [`issues_edit_issue_job.rb`](https://github.com/huboard/huboard-web/blob/adf9aa173743c0bcc36dee004796ccca95991811/app/jobs/api/issues_edit_issue_job.rb). Is there a base job they could inherit from?

Also, what's relationship between this new job and [`issues_assign_milestone_job.rb`](https://github.com/huboard/huboard-web/blob/adf9aa173743c0bcc36dee004796ccca95991811/app/jobs/api/issues_assign_milestone_job.rb)?

rauhryan:
@dahlbyk yeah,  you are right. 

I looked into it a little closer

I thought I could just shovel the event to a generic `edited` handler but we don't handle that event client side yet. 

So, I thought I'd future proof it a little bit. 

Added a method to the job resolver to t conventionally resolve webhook event jobs and if one can't be found it will fallback to the `issues_edit_issue_job`

Should be generic enough to catch things if new events pop up in the future

PR #388, opened 2016-11-15
Compound health checks

rauhryan:


discorick:
One small thing:
image

but still seeing: 
image

Looks like the alert isn't picking up

PR #389, opened 2016-11-23
Reports the most recent commit status on PR's

discorick:
Small logic tweak, the github API returns commit shas from oldest -> newest, which we were using to fetch statuses. Now using the most recent commit sha to fetch the status.


rauhryan:
🚢

PR #390, opened 2016-12-07
Changes the sorting logic of the assignees selector

rauhryan:
Mimics the behavior of the github UI selector

closes huboard/huboard#698 when merged

### DIFF
--- a/ember-app/app/components/hb-milestone.js
+++ b/ember-app/app/components/hb-milestone.js
@@ -9,32 +9,42 @@ var HbMilestoneComponent = Ember.Component.extend({
   isOpen: function(){
     return false;
   }.property(),
-
   listItems: function () {
-
     return this.get("milestones")
     .filter(function(item) {
       var term = this.get("filterMilestones") || "";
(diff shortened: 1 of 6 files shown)


ember-app/app/components/hb-milestone.js
```diff
@@ -9,32 +9,42 @@ var HbMilestoneComponent = Ember.Component.extend({
...
+      if(b.data.title === this.get('selected.title')) {
+        return 1;
+      }
 
+      return a.data.title.localeCompare(b.data.title);
+    }.bind(this))
+    .map(function(item) {
       return this.ListItem.create({
-        selected: item.data.id === this.get("selected.id"),
+        selector: this,
```
discorick: Curious what prompted this change?

rauhryan: the list item is mapped off the repo milestones, but only one can be selected because it's a single property. 

item.selected was being mapped off of the selectors selected property, but I don't want that to immediately cause a recomputed of the list items, UX wise is weird to have the milestone selected jump to the top as soon as you click it...

discorick: Ah, that makes sense - especially if dealing with a long list of milestones 👍

PR #395, opened 2017-01-20
Troubleshoot heroku issues

rauhryan:


discorick:
Combined with the Uglifier bump this seems to fix all the current PR's

PR #396, opened 2017-01-20
Added fix for issue 703 to make checkmark only show on selected users

ImpulseChimp:
I hesitate to say fix since I find it more of a UI preference, but I removed the on hover check mark for assigned users on cards since once deselected, it can be confusing since they still look selected until you hover off. I opened issue 703 at [the parent repo](github.com/huboard/huboard) for this. 

Part of my reasoning for this is that the purple hover color is also a great indicator of who you can select so adding the check mark on top of that seems unnecessary. That is why I removed it completely instead of just removing it upon deselect. (I also removed some white space since I'm that kinda guy...)

Before
beforeissue703dev

After
afterissue703

Hopefully this fits with the UI design/usability that you are aiming for.

### DIFF
--- a/app/assets/stylesheets/components/_selector.scss
+++ b/app/assets/stylesheets/components/_selector.scss
@@ -5,7 +5,7 @@
 
 }
 .hb-selector-component {
-  
+
   margin-top: 12px;
   padding-top: 12px;
   border-top: 1px solid #ccc;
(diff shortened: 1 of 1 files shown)


discorick:
Awesome thanks for the PR! Would you mind pushing your code straight the to huboard/huboard-web remote to trigger CI and auto-deploy ? As soon as that is done I will take a look and let you know what I think. 

Additionally theres a deployment fix I merged into master this weekend, could you rebase that into your work ?

